PR #38, opened 2022-09-13
Change dapt: prefix to daptm:

nigelmegitt:
Closes #22.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/38.html" title="Last updated on Sep 13, 2022, 12:45 AM UTC (71455ca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/38/4c5d1c1...71455ca.html" title="Last updated on Sep 13, 2022, 12:45 AM UTC (71455ca)">Diff</a>

PR #39, opened 2022-09-13
Bring in content from ADPT

### LINKED ISSUE #22: Namespace prefix `dapt` should not map to a metadata fragment id
If we need a metadata only prefix then we should include something to hint at that, such as an `m`, e.g. `daptm`. That leaves us clear to define any DAPT-specific vocabulary in a `dapt` prefixed namespace without the `#metadata` fragment ID if we ever need to.

### LINKED ISSUE #23: Forward compat: rules for unrecognised vocabulary
The section on namespace says:
>Implementations shall ignore elements and attributes in the DAPT namespaces that they do not recognize.
>
>Implementations shall ignore elements and attributes in non TTML-namespaces that they do not recognize.

We should consider if _ignore_ is the right thing here. Should we require that they attempt to preserve any such content without processing it?

For example, in my hypothetical document I have:

```xml
<p foreignNamespace:importantAttribute="forgetmenot">
   <foreignNamespace:importantElement>My workflow depends on this</foreignNamespace:importantElement>
   <span ...>My normal DAPT content here</span>
</p>
```

An implementation that obeys the SHALL requirement, on processing such a document, e.g. parsing it, doing something and then reserialising it as XML, would strip out the foreignNamespace attributes and elements.

We should at the very least permit the implementation to preserve that content.

### LINKED ISSUE #26: Add profile constraints expressed as features
Since we are defining a profile of TTML2 we need to list the permitted/required/prohibited/optional features from TTML2 and those defined locally in DAPT in the conformance constraints. For example see https://w3c.github.io/adpt/#features-section and https://www.w3.org/TR/ttml-imsc1.2/#supported-features-and-extensions

nigelmegitt:
Closes #26 as a first pass - probably additional changes will also be needed to the profile constraints. Also closes #23 by using hopefully better wording for the handling of foreign vocabulary. Closes #22 by using DAPT Metadata `daptm:` namespace prefix.

Modifies ADPT conformance by additionally defining the disposition "required" for features, and using it for `#contentProfiles`.

Brings in:
* localBiblio
* Documentation Conventions
* Definitions
* Conformance
* Profile Resolution Semantics
* Additional namespaces used within the specification (maybe too many)
* Related Video Object
* Synchronization requirements
* Reorganised profile signalling
* List of features with dispositions for each (this almost certainly will need further edits)

Also replaces SHALL and shall with MUST to be more "modern".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/39.html" title="Last updated on Sep 14, 2022, 6:06 PM UTC (f278bbf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/39/41b63ac...f278bbf.html" title="Last updated on Sep 14, 2022, 6:06 PM UTC (f278bbf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -172,6 +172,9 @@ table.coldividers td + td { border-left:1px solid gray; }
         <dfn>Audio description</dfn>	An audio rendition of a <a>Description</a> or a set of <a>Descriptions</a>.
       </p>
       <p>
+        <dfn>Audio description mixed audio track</dfn>  The output of an audio mixer incorporating the main programme audio and the audio description.
+      </p>
+      <p>
         <dfn data-lt="Description|Descriptions">Description</dfn>  A set of words that describe an aspect of the programme presentation, suitable for rendering into audio by means of vocalisation and recording or used as a <a>Text Alternative</a> source for text to speech translation.
       </p>
       <p>
@@ -191,6 +194,9 @@ table.coldividers td + td { border-left:1px solid gray; }
       </p>
       <p>
         <dfn data-cite="ttml2#style-value-lwsp">Linear White-Space</dfn> See Section 2.3 at [[TTML2]].
+      </p>
+      <p>
+        <dfn>Main programme audio</dfn> The audio associated with the programme prior to any mixing with audio description.
       </p>
       <p>
         <dfn data-cite="ttml2#terms-presentation-processor">Presentation processor</dfn> See Section 2.2 at [[TTML2]].
@@ -388,8 +394,8 @@ table.coldividers td + td { border-left:1px solid gray; }
         </section>
 
         <section>
-          <h4>Description</h4>
-          <p>Each <a>Event</a> MAY have a <dfn>Description</dfn> property, as a human-readable description of the <a>Event</a>. </p>
+          <h4>Event Description</h4>
+          <p>Each <a>Event</a> MAY have an <dfn>Event Description</dfn> property, as a human-readable description of the <a>Event</a>. </p>
           <p class="issue" data-number="30"></p>
           <p class="note">It does not have to be unique for this <a>Event</a>. It may be used to identify in a human-readable way one or more <a>Events</a> that need to be processed together, e.g. in a batch recording.</p>
           <p class="issue" data-number="20"></p>
--- a/index.html
+++ b/index.html
@@ -108,7 +108,8 @@ table.coldividers td + td { border-left:1px solid gray; }
       <h2>Scope</h2>
       <p>This specification defines a text-based profile of the Timed Text Markup Language version 2.0 [[TTML2]] based on [[ttml-imsc1.1]].
         This profile is intended to support dubbing and audio description workflows worldwide,
-        and to meet the requirements defined in [[[?DAPT-REQS]]].
+        to meet the requirements defined in [[[?DAPT-REQS]]], and to permit the visual presentation
+        features within [[ttml-imsc1.1]].
       <p class="issue" data-number="5"></p>
     </section>
 
@@ -159,7 +160,7 @@ table.coldividers td + td { border-left:1px solid gray; }
         <dfn>Audio description</dfn>	An audio rendition of a <a>Description</a> or a set of <a>Descriptions</a>.
       </p>
       <p>
-        <dfn data-lt="Description|Descriptions">Description</dfn>  A set of words that describe an aspect of the programme presentation, suitable for rendering into audio by means of vocalisation and recording or used as a <a>Text Alternative</a> source for speech to text translation.
+        <dfn data-lt="Description|Descriptions">Description</dfn>  A set of words that describe an aspect of the programme presentation, suitable for rendering into audio by means of vocalisation and recording or used as a <a>Text Alternative</a> source for text to speech translation.
       </p>
       <p>
         <dfn data-cite="ttml2#terms-default-region">Default Region</dfn> See Section 11.3.1.1 at [][TTML2]].
@@ -434,18 +435,19 @@ table.coldividers td + td { border-left:1px solid gray; }
       <section id="foreign-elements-and-attributes">
         <h3>Foreign Element and Attributes</h3>
         <p>
-          A Document Instance MAY contain elements and attributes that are neither specifically permitted nor forbidden by a profile.
-        </p>
-        <p>
-          A transformation processor SHOULD preserve such elements or attributes whenever possible.
+          A <a>Document Instance</a> MAY contain elements and attributes that are neither specifically permitted nor forbidden by a profile.
         </p>
         <p class ="note">
-          Document Instances remain subject to the content conformance requirements specified at Section 3.1 of [[TTML2]].
+          <a>Document Instances</a> remain subject to the content conformance requirements specified at Section 3.1 of [[TTML2]].
           In particular, a Document Instance can contain elements and attributes not in any TT namespace, i.e. in foreign namespaces, since such elements and attributes are pruned by the algorithm at Section 4 of [[TTML2]] prior to evaluating content conformance.
         </p>
         <p class ="note">
           For validation purposes it is good practice to define and use a content specification for all foreign namespace elements and attributes used within a Document Instance.
         </p>
+        <p>
+          A <a>transformation processor</a> SHOULD preserve such elements or attributes whenever possible.
+        </p>
+        <p class="ednote">Do we need to say that a presentation processor may ignore foreign vocab?</p>
       </section>
       <section id="namespaces">
         <h3>Namespaces</h3>
@@ -510,16 +512,21 @@ table.coldividers td + td { border-left:1px solid gray; }
       <section id="related-video-object-section">
         <h3>Related Video Object</h3>
         <p>
-          A Document Instance MAY be associated with a <a>Related Video Object</a>.
+          A <a>Document Instance</a> MAY be associated with a <a>Related Video Object</a>.
         </p>
         <p class="note">
-          While this specification contains specific provisions when a Document Instance is associated with a <a>Related Video Object</a>, it does not prevent the use of a Document Instance with other kinds of Related Media Object, e.g. an audio only object.
+          While this specification contains specific provisions when a <a>Document Instance</a> is associated with a <a>Related Video Object</a>, it does not prevent the use of a Document Instance with other kinds of Related Media Object, e.g. an audio only object.
         </p>
       </section>
       <section id="synchronization-section">
         <h3>Synchronization</h3>
+        <p class="ednote">Check if we need this, and if we do, rework for audio - e.g. relate to audio sample, or other quantisation units?</p>
         <p>
-          Each intermediate synchronic document of the Document Instance is intended to be presented (audible) starting on a specific frame and removed (inaudible) by a specific frame of the Related Video Object.
+          Each intermediate synchronic document of the Document Instance is intended to be rendered starting on a specific frame and removed by a specific frame of the Related Video Object.
+        </p>
+        <p class="note">In the context of this specification rendering could be visual presentation of text,
+          for example to show an actor what words to speak, or could be audible playback of an audio resource,
+          or could be physical or haptic, such as a Braille display.
         </p>
         <p>
           When mapping a media time expression M to a frame F of a Related Video Object (or Related Media Object), e.g. for the purpose of mixing audio sources signalled by a Document Instance into the main program audio of the <a>Related Video Object</a>, the presentation processor MUST map M to the frame F with the presentation time that is the closest to, but not less, than M.
@@ -536,9 +543,6 @@ table.coldividers td + td { border-left:1px solid gray; }
       </section>
       <section id="profile-signaling-section">
         <h3>Profile Signaling</h3>
-        <p>
-          The <code>ttp:profile</code> attribute SHOULD be present on the <code>tt</code> element and equal to the designator of the DAPT 1.0 profile to which the Document Instance conforms.
-        </p>
         <p>TTML documents representing DAPT Scripts MUST specify a <code>ttp:contentProfiles</code> attribute
           on the <code>tt</code> element with one value equal to the designator of the <a>DAPT 1.0</a> profile
           to which the Document Instance conforms.
@@ -570,6 +574,11 @@ table.coldividers td + td { border-left:1px solid gray; }
         <p>
           See <a href="#conformance" class="sec-ref">Conformance</a> for a definition of <em>permitted</em>, <em>prohibited</em> and <em>optional</em>.
         </p>
+        <p class="ednote">Intent is to make it as easy as possible to transform a Document Instance into
+          IMSC, so we need to check that there are no IMSC permitted features that we prohibit, or if there
+          are, then we can explain the reason.
+        </p>
+        <p class="ednote">Editorial task: go through this list of features and check the disposition of each. IMSC-only features should be optional.</p>
         <table class="simple">
           <tbody>
             <tr>
@@ -1145,7 +1154,7 @@ table.coldividers td + td { border-left:1px solid gray; }
                 <li>The following attribute corresponding to the <a>On Screen</a> <a>Event</a> property may be present:
                   <div class="exampleInner">
                     <pre class="nohighlight">
-dapt:onScreen
+daptm:onScreen
   : "ON" 
   | "OFF"
   | "ON_OFF"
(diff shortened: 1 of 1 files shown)


nigelmegitt:
@cconcolato given https://github.com/w3c/dapt/pull/38#issuecomment-1245876257 please could you approve this so we can merge it? May need further changes afterwards of course.

index.html
```diff
@@ -26,7 +26,16 @@
       github: "w3c/dapt",
       wgPublicList: "public-tt",
 
-      };
+      // local bibliography
+      localBiblio: {
+          "WHP051": {
```
cconcolato, line 31: This reference does not seem used.

index.html
```diff
@@ -120,7 +130,128 @@ <h2>Introduction</h2>
...
-  <section id="conformance"></section>
+    <section id="conventions">
+      <h2>Documentation Conventions</h2>
+      <p>
+        This document uses the same conventions as [[TTML2]] for the specification of parameter attributes, styling attributes and metadata elements. In particular:
+      </p>
+      <p>
+        Section 2.3 of [[TTML2]] specifies conventions used in the [[XML]] representation of elements; and
+        Sections 6.2 and 8.2 of [[TTML2]] specify conventions used when specifying the syntax of attribute values.
+      </p>
```
cconcolato, line 141: The TTML2 conventions are large and not all needed. I'm worried about making this spec unnecessarily hard to read. Can we rather pull in only the conventions we need? 
For example the "may"/"should"/must" part is already covered in the conformance section. Another example is the use of "deprecated"/"obsolete" that are not (yet) needed.

index.html
```diff
@@ -120,7 +130,128 @@ <h2>Introduction</h2>
...
+      <p>
+        All content of this specification that is not explicitly marked as non-normative is considered to be normative.
+        If a section or appendix header contains the expression "non-normative", then the entirety of the section or appendix is considered non-normative.
+      </p>
+      <p>
+        This specification uses Feature designations as defined in Appendices E at [[TTML2]]:
+        when making reference to content conformance, these designations refer to the syntactic expression or the semantic capability associated with each designated Feature; and
+        when making reference to processor conformance, these designations refer to processing requirements associated with each designated Feature.
+        If the name of an element referenced in this specification is not namespace qualified, then the TT namespace applies (see <a href="#namespaces">9.3 Namespaces</a>.)
+      </p>
```
cconcolato, line 151: Is this really a "document convention"? It would seem more appropriate in the conformance section.

index.html
nigelmegitt: OK, will bring in relevant description of AD that references it.

index.html
nigelmegitt: I think it is clear enough that only a subset of the conventions are being brought in.

index.html
```diff
@@ -120,7 +130,128 @@ <h2>Introduction</h2>
...
+      <p>
+        <dfn>Audio description</dfn>	An audio rendition of a <a>Description</a> or a set of <a>Descriptions</a>.
+      </p>
+      <p>
+        <dfn data-lt="Description|Descriptions">Description</dfn>  A set of words that describe an aspect of the programme presentation, suitable for rendering into audio by means of vocalisation and recording or used as a <a>Text Alternative</a> source for text to speech translation.
+      </p>
+      <p>
+        <dfn data-cite="ttml2#terms-default-region">Default Region</dfn> See Section 11.3.1.1 at [][TTML2]].
+      </p>
```
cconcolato, line 167: This definition is not used, as expected since we don't want to specify layout. Can we remove it?

index.html
nigelmegitt: I'm not sure it makes much practical difference?

index.html
nigelmegitt: Yes

index.html
```diff
@@ -120,7 +130,128 @@ <h2>Introduction</h2>
...
+      <p>
+        <dfn data-lt="Document Instance|Document Instances" data-cite="ttml2#terms-document-instance">Document Instance</dfn> As defined by [[TTML2]].
+      </p>
+      <p>
+        <dfn data-cite="ttml2#terms-document-interchange-context">Document Interchange Context</dfn> As defined by [[TTML2]].
+      </p>
+      <p>
+        <dfn data-cite="ttml2#terms-document-processing-context">Document Processing Context</dfn>  See Section 2.2 at [[TTML2]].
+      </p>
```
cconcolato, line 176: Can we remove these definitions? They are used only in the "Profile Resolution Semantics" and are quite complex concepts that I wish we could shield the DAPT reader/implementor from.

index.html
cconcolato: It is not clear to me.

index.html
cconcolato: It does I think since we can then have a discussion on conformance in one place.

index.html
```diff
@@ -114,13 +124,143 @@ <h2>Introduction</h2>
...
+      The use of the terms <a>presentation processor</a> 
+      and <a>transformation processor</a> within this document does not imply conformance <i>per se</i> to any of the Standard Profiles defined in [[TTML2]].
+      In other words, it is not considered an error for a <a>presentation processor</a> or <a>transformation processor</a> to conform to the profile defined in this document without also conforming to the TTML2  Presentation Profile or the TTML2 Transformation Profile.
+    </p>
+    <p class="note">
+      This document does not specify presentation processor or transformation processor behavior when processing or transforming a non-conformant Document Instance.
+    </p>
+    <p class="note">
+      The permitted and prohibited dispositions do not refer to the specification of a ttp:feature or ttp:extension element as being permitted or prohibited within a ttp:profile element.
+    </p>
```
cconcolato, line 268: Is this note really needed? Can avoid having the reader/implementer worry about `ttp:feature` et al.

index.html
```diff
@@ -114,13 +124,143 @@ <h2>Introduction</h2>
...
+      <li>MUST NOT include any vocabulary, syntax or attribute value associated with a Feature whose disposition is prohibited in the profile.</li>
+    </ul>
+    <p class="note">
+      A <a>Document Instance</a>, by definition, satisfies the requirements of Section 3.1 at [[TTML2]], and hence a Document Instance that conforms to a profile defined herein is also a conforming TTML2 Document Instance.
+    </p>
+    <p>
+      A <a>presentation processor</a> that conforms to the profile defined in this specification:
+    </p>
+    <ul style="list-style-type:disc">
+      <li>MUST satisfy the Generic Processor Conformance requirements at Section 3.2.1 of [[TTML2]]</li>
```
cconcolato, line 244: Do I understand correctly that it is normal than a "presentation processor" does not satisfy the requirements in 3.2.3?  and similarly for "transformation processor" and section 3.2.2.

index.html
nigelmegitt: I think it's almost in the boilerplate category, and we need it for specification clarity. I don't think we should worry about it. But we should reference those terms properly in the Profile Resolution Semantics section.

index.html
nigelmegitt: I think this is an example of retaining lessons learned in the past. I'd keep it. It avoids a potential ambiguity.

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+        </p>
+        <p>
+          If the resolved profile is not the Profile supported by the Processor but is feasibly interoperable with the Profile, then the resolved profile is the Profile.
+          If the resolved profile is undetermined or not supported by the Processor, then the Processor SHOULD nevertheless process the Document Instance using the Profile; otherwise, processing MAY be aborted.
+          If the resolved profile is not the proposed Profile, processing is outside the scope of this specification.
+        </p>
+        <p>
+          If the resolved profile is the profile supported by the Processor, then the Processor SHOULD process the Document Instance according to the Profile.
+        </p>
+      </section>
```
cconcolato, line 440: Do we need this whole section given that we require setting `ttp:contentProfiles`?

index.html
nigelmegitt: Yes, I think that's right. There is even a note about it!

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+      <section id="Document Encoding">
+        <h3>Document Encoding</h3>
+        <p>
+          A Document Instance MUST use UTF-8 character encoding as specified in [[UNICODE]].
+        </p>
+      </section>
+      <section id="foreign-elements-and-attributes">
+        <h3>Foreign Element and Attributes</h3>
+        <p>
+          A <a>Document Instance</a> MAY contain elements and attributes that are neither specifically permitted nor forbidden by a profile.
```
cconcolato, line 453: > by a profile
Can we rephrase and say by this specification?

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+        </p>
+        <p class ="note">
+          <a>Document Instances</a> remain subject to the content conformance requirements specified at Section 3.1 of [[TTML2]].
+          In particular, a Document Instance can contain elements and attributes not in any TT namespace, i.e. in foreign namespaces, since such elements and attributes are pruned by the algorithm at Section 4 of [[TTML2]] prior to evaluating content conformance.
+        </p>
+        <p class ="note">
+          For validation purposes it is good practice to define and use a content specification for all foreign namespace elements and attributes used within a Document Instance.
+        </p>
```
cconcolato, line 461: Why do we need this note?

index.html
```diff
@@ -114,13 +124,143 @@ <h2>Introduction</h2>
...
+      </p>
+      <p>
+        <dfn data-cite="ttml2#style-value-lwsp">Linear White-Space</dfn> See Section 2.3 at [[TTML2]].
+      </p>
+      <p>
+        <dfn data-cite="ttml2#terms-presentation-processor">Presentation processor</dfn> See Section 2.2 at [[TTML2]].
+      </p>
+      <p>
+        <dfn data-cite="ttml2#terms-processor">Processor</dfn>  Either a Presentation processor or a Transformation processor.
+      </p>
```
cconcolato, line 206: I would prefer if we defined one (or more) classes of DAPT Processors, for example DAPT Authoring Tool, DAPT Audio Recorder/Renderer, DAPT-to-IMSC Convertor. For each of them, we could say if they are "TTML2 presentation processors" or "TTML2 transformation processors" for those who are interested in digging into what that means, but in practice a DAPT implementor should not need to care about it.

index.html
nigelmegitt: Good question - suggest we raise an issue for that and handle via a separate PR.

index.html
nigelmegitt: Yes if we are sure we're only defining one specification.

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+        </p>
+        <p>
+          The namespaces defined by this proposal document are mutable [[namespaceState]]; all undefined names in these namespaces are reserved for future standardization by the W3C.
+        </p>
+      </section>
+      <section id="related-video-object-section">
+        <h3>Related Video Object</h3>
+        <p>
+          A <a>Document Instance</a> MAY be associated with a <a>Related Video Object</a>.
+        </p>
```
cconcolato, line 531: We should update the "MAY" to "is". In DAPT workflows, as illustrated in the DAPT Requirements diagram, it always start with an Audio/Video Programme.

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+        <p class="note">
+          While this specification contains specific provisions when a <a>Document Instance</a> is associated with a <a>Related Video Object</a>, it does not prevent the use of a Document Instance with other kinds of Related Media Object, e.g. an audio only object.
+        </p>
```
cconcolato, line 534: Can we remove this note?

index.html
nigelmegitt: It's advice that if you're going to add extension metadata you should specify it properly. Just trying to encourage good practice.

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+        <p>
+          EXAMPLE 1
+          A media time expression of 00:00:05.1 corresponds to frame ceiling( 5.1 × ( 1000 / 1001 × 30) ) = 153 of a <a>Related Video Object</a> with a frame rate of 1000 / 1001 × 30 ≈ 29.97.
+        </p>
+        <p class="note">
+          In typical scenario, the same video program (the <a>Related Video Object</a>) will be used for Document Instance authoring, delivery and user playback.
+          The mapping from media time expression to <a>Related Video Object</a> above allows the author to precisely associate <a>audio description</a> content with video frames, e.g. around existing audio dialogue and sound effects.
+          In circumstances where the video program is downsampled during delivery, the application can specify that, at playback, the relative video object be considered the delivered video program upsampled to is original rate, thereby allowing audio content to be presented at the same temporal locations it was authored.
+        </p>
+      </section>
```
cconcolato, line 558: Can we remove the entire section?

index.html
nigelmegitt: I'm happy to do that if it's easier to read - I suggest we open a separate issue and PR to make that change. The formal specification is important, but we should promote readability.

index.html
```diff
@@ -278,47 +418,665 @@ <h4>Identification and Profile Signaling</h4>
...
+                <td><dfn data-dfn-type="dfn" id="dfn-dapt-1-0">DAPT 1.0</dfn></td>
+    
+                <td><code>http://www.w3.org/ns/ttml/profile/dapt</code></td>
+              </tr>
+            </tbody>
+          </table>
+        </section>
+      </section>
+      <section id="features-section">
+        <h3>Features</h3>
```
cconcolato, line 588: I would suggest moving the mapping of DAPT to TTML2 feature to an annex. The DAPT/Reader implementor needs to understand the Features of DAPT, not necessarily how they map to TTML2.

index.html
nigelmegitt: I think I disagree - there are cases where a processor can transform a DAPT document without any specific knowledge of the related video; also we don't require any metadata connecting to a related video. At least in the BBC there may be cases where the same document is associated with _multiple_ videos.

I think we need to look at this whole section via a separate issue though - the whole link to video frames for visual presentation seems irrelevant.

index.html
nigelmegitt: Again, I think an issue to consider the relationship to some other audio or audiovisual media object should include this in scope.

index.html
nigelmegitt: I suggest leaving it and opening an issue to fix it, because it's clearly broken. Or delete it and open an issue to define synchronisation requirements. Happy either way.

index.html
nigelmegitt: Happy to see how that looks via a separate PR. Also worth noting (I think there's an Ed Note) that the dispositions are likely to be wrong in many cases, since we need to make IMSC features "optional".

index.html
cconcolato: I find it too hard to understand for the non expert. People will do whatever they want anyway with their extensions and validation tools. I think we should get rid of the note.

cconcolato (approved):
Approving the PR to ease the editing process, but we should keep the conversations going and act in additional PRs.

index.html
cconcolato: done https://github.com/w3c/dapt/issues/43

index.html
cconcolato: done https://github.com/w3c/dapt/issues/44

index.html
cconcolato: This is related to https://github.com/w3c/dapt/issues/43

index.html
cconcolato: We should open an issue to discuss whether we need one profile or two profiles (audio-support or not).

index.html
cconcolato: I opened https://github.com/w3c/dapt/issues/45. I would prefer deleting the section.

index.html
cconcolato: https://github.com/w3c/dapt/issues/46

index.html
cconcolato: I opened https://github.com/w3c/dapt/issues/45

index.html
cconcolato: https://github.com/w3c/dapt/issues/45

index.html
cconcolato: https://github.com/w3c/dapt/issues/47

Issue #21, opened 2022-09-12
Rename Contextual Text to "Translated Text"


nigelmegitt:
The data model term "Contextual Text" suggests that it is text about context, but is defined as text in another language from the original, which would be better called "Translated Text".

cconcolato:
It is not necessarily translated. It could be the original text that is preserved for context.

nigelmegitt:
OK, in that case the data model description is unclear. We could also consider something like "Other Language Text"?

cconcolato:
We can certainly improve the description but I still prefer "Context" to "Other Language".

cconcolato:
Related to https://github.com/w3c/dapt/issues/73

nigelmegitt:
Editors call - agreed that the definition of what is "main text" vs "contextual text" should not change with the workflow step, but should stay consistent as the authoring process progresses. The interpretation of any particular text as being "main" or "contextual" is in the eye of the beholder, in that it depends what the person working on the document needs to deliver in that phase of the workflow.

Instead, we should label original language text and translated text as such, and the person working on the document would know what their deliverable is, one or the other.

The question then is what to call these in the data model. Proposal is not to label them at all. Remove "Main text" and "Contextual text" and just have "Text", cardinality 0..*, with examples.

Examples would include:
1. Original Language Dialog List containing original language transcription text only.
2. Translated Dialog List contains translation text (and possibly also a 2nd translation text for the pivot language) and optionally the original language transcription text as well.

Any constraints on the contents of the document for each Script Type can be explained in a separate section.

nigelmegitt:
I think #77 resolves this issue, not exactly by renaming Contextual Text, but by removing the concept altogether, in favour of markup on each Text object saying whether it is in the Original language or Translated.

PR #77, opened 2022-11-28
Define original vs translation independently

### LINKED ISSUE #73: Contextual text is inadequately identified
In the current ED, it is unclear which Text is a translation and which is original language. In https://w3c.github.io/dapt/#text there is a Note that says that the `xml:lang`, being different to the primary language (`xml:lang` on `tt` element) can be used, but that does not work when the original language dialogue contains multiple languages (especially if one of those languages is the target language for the translation).

I suggest that Translated Dialogue List should specify the target language(s), and each Text entity should be able to carry additional metadata signifying if it is a translation, or if it is the same language as the original programme dialogue.

This allows for any sections of dialogue that are in the target language of the translation to be retained and flagged as "no need to dub", and reused for hard of hearing subtitles if needed downstream.

nigelmegitt:
Resolves #73 and #21.

* Allows the content to specify using a new Text Language Source property if it is in the original language or is a translation independently of any logic around `xml:lang` on the `<tt>` element and `xml:lang` defined more locally, which resolves some otherwise ambiguous scenarios.
* Add informative text on the markup of the expected content for each Script Type.
* Fix up the introductory examples to include the new syntax
* Explain the relationship between Primary Language and Text Language Source
* Remove Contextual and Main text concepts
* Simplify the constraints on Script Events containing Text objects
* Require Text objects to specify the Text Language Source
* Relax the MUST constraint on `p` elements setting `xml:lang` to a SHOULD, and add an explanatory note


index.html
```diff
@@ -352,17 +352,64 @@ <h4>Script Type</h4>
...
             <p>The definitions of the types of documents and the corresponding <code>ttm:role</code> values are:
               <ul>
-                <li><dfn>Original Language Dialogue List</dfn>: When the <code>ttm:role</code> value is <code>x-DUBBING_ORIGINAL_DIALOGUE_LIST</code>, the document represents a literal transcription of the dialogue and on-screen text in their original spoken/written language(s).
-                <aside class="example">If a movie contains dialogue in English and Hebrew, the <a>Original Language Dialogue List</a> will contain some <a>Script Events</a> in English and some in Hebrew. None of the <a>Script Events</a> is translated.</aside>
+                <li>
+                  <dfn>Original Language Dialogue List</dfn>:
+                  When the <code>ttm:role</code> value is <code>x-DUBBING_ORIGINAL_DIALOGUE_LIST</code>,
+                  the document represents a literal transcription of the dialogue and on-screen text in their original spoken/written language(s).
+                <aside class="note"><a>Text</a> objects in the Original Language Dialogue List have their <a>Text Language Source</a> set to <a>Original</a>.</aside>
```
cconcolato, line 359: I understand we don't want to have a normative statement in the note, but this makes the sentence ambiguous. Could we have a forward reference to where the actual normative statement is? Something like, "As specified in XXX, Text objects have their TLS set to Original"?

index.html
```diff
@@ -352,17 +352,64 @@ <h4>Script Type</h4>
...
+                <aside class="example">If a movie contains dialogue in English and Hebrew,
+                  the <a>Original Language Dialogue List</a> will contain some <a>Script Events</a> in English and some in Hebrew,
+                  all marked as <a>Original</a> <a>Text Language Source</a>.
+                  None of the <a>Script Events</a> is marked as <a>Translation</a>.</aside>
```
cconcolato, line 363: Similar comment. We should make sure that these notes are expressed elsewhere as normative statements.

index.html
```diff
@@ -352,17 +352,64 @@ <h4>Script Type</h4>
...
-                <aside class="example">If a movie contains dialogue in English and Hebrew, the French <a>Translated Dialogue List</a> will contain at least the translation in French of all <a>Script Events</a>. It may still retain text content in Hebrew and English to assist further processing.</aside>
+                <li>
+                  <dfn>Translated Dialogue List</dfn>:
+                  When the <code>ttm:role</code> value is <code>x-DUBBING_TRANSLATED_DIALOGUE_LIST</code>,
+                  the document represents a translation of the dialogue or on-screen text in a common language.
+                  It is meant to be adapted to produce a <a>Pre-Recording Dub Script</a>. 
+                <aside class="note">Translated <a>Script Events</a> contain <a>Text</a> objects whose
+                  <a>Text Language Source</a> is set to <a>Translation</a>.
+                  They can also contain <a>Original</a> language <a>Text</a>.</aside>
+                <aside class="note">A <a>Script Event</a> can contain <a>Translation</a> <a>Text</a> objects in more than one language, for context,
```
cconcolato, line 373: I think this should say:
```
A <a>Script Event</a> in this type of script has a <a>Translation</a> <a>Text</a> object in the target recording language but it may contain an <a>Original</a> <a>Text</a> object, for context, to assist in the adaption process.</aside>
```

index.html
```diff
@@ -530,27 +590,65 @@ <h3>Script Event</h3>
...
         </section>
 
+        <section>
+          <h4>Text Language Source</h4>
+          <p>The <dfn>Text Language Source</dfn> property is an annotation indicating whether a <a>Text</a> object is
+            in the same language as the relevant part of the <a>Related Media Object</a>'s
+            language (original), or if it is a representation in another language (translation):</p>
+          <ul>
+            <li><dfn>Original</dfn> - the <a>Text</a> is in the same language as the dialogue that it represents in the original programme audio,
+              or it represents non-dialogue sound and is in the <a>Primary Language</a> of the <a>Document Instance</a>.</li>
```
cconcolato, line 628: Non-dialog sounds don't have a language, do they? Did you mean "on screen text" but in that case, it may also not be in primary language.

index.html
```diff
@@ -530,27 +590,65 @@ <h3>Script Event</h3>
...
+            <li><dfn>Translation</dfn> - the <a>Text</a> is a representation of the Original language text in a different language.</li>
+          </ul>
+          <aside class="note">The language of a <a>Text</a> object whose <a>Text Language Source</a> is <a>Original</a>
+            can be different to the document's <a>Primary Language</a>.
+            For example a transcript of a short piece of dialogue spoken in a different language to the majority of the dialogue would be annotated
+            with both a different <code>xml:lang</code> value and as being <a>Original</a>.</aside>
+          <aside class="note"><a>Original</a> text is known as "Main",
+            and <a>Translation</a> text as "Contextual", in some workflows.</aside>
```
cconcolato, line 636: We probably can get rid of this note.

index.html
```diff
@@ -530,27 +590,65 @@ <h3>Script Event</h3>
...
+          <p>The <a>Text Language Source</a> property is represented as a <code>daptm:tls</code> attribute with the following constraints:</p>
+          <p class="ednote">Should we use an abbreviated attribute name?</p>
```
cconcolato, line 638: maybe simply `dapt:languageSource`?

index.html
nigelmegitt: Yes, text representations of non-dialogue sounds have a language. If the text is "clapping" that's an English language description of a non-dialogue sound. Agreed the sound itself doesn't have a language, but it's unusual to try to write the sound down phonetically.

index.html
nigelmegitt: OK, that works - my only query is whether it makes sense to make frequently used attribute names shorter, but I think semantic clarity is more important.

index.html
nigelmegitt: Would we need to define "target recording language"?

index.html
nigelmegitt: We don't have any such normative statements so far. I'm not sure what we could state, normatively?

index.html
cconcolato: Oh, I see. Can we also add on-screen too?

index.html
cconcolato: `langSrc`?

nigelmegitt: Done in 3cfe8db.

index.html
nigelmegitt: Done in 3cfe8db.

index.html
nigelmegitt: Done in 3cfe8db. Please review this change.

index.html
nigelmegitt: Done in 3cfe8db.

index.html
nigelmegitt: Made an attempt to do this in 3cfe8db - please review. I could only come up with SHOULD / SHOULD NOT and MAY statements, but they're still probably clearer than the notes.

cconcolato (approved):

Issue #57, opened 2022-09-22
[ED issue] Consider using the term document vs document instance


cconcolato:
We should use the term "document" (meaning an XML Document compliant to TTML) rather than "Document Instance" which is defined in TTML:
> A concrete realization of a [timed text markup language document](https://www.w3.org/TR/2018/REC-ttml2-20181108/#terms-timed-text-markup-language-document), where the concrete form is specific to the context of reference.

with "timed text markup language document" being defined as:
> An abstract document that is purported or confirmed to be a [valid abstract document instance](https://www.w3.org/TR/2018/REC-ttml2-20181108/#terms-valid-abstract-document-instance).

and "valid abstract document instance" being defined as:
> An [abstract document instance](https://www.w3.org/TR/2018/REC-ttml2-20181108/#terms-abstract-document-instance) which has been assessed for validity and found to be valid as defined by [4 Document Types](https://www.w3.org/TR/2018/REC-ttml2-20181108/#doctypes).

and "abstract document instance" being defined as:
> An instance of an abstract data set as represented by a [reduced xml infoset](https://www.w3.org/TR/2018/REC-ttml2-20181108/#terms-reduced-xml-infoset).

nigelmegitt:
I don't see the difference between "document == an XML Document compliant to TTML" and "Document Instance" other than that the latter is more clearly defined, in the end.

I'd suggest that we define "document" with an easily understood non-normative meaning, and a normative meaning that is "A Document Instance conformant to the profile defined in this specification", where Document Instance is as per TTML2.

PR #79, opened 2022-11-29
Specify entity reference requirements

### LINKED ISSUE #51: Make concrete the suggested requirement from TTML2 P.1
[TTML2 P.1](https://www.w3.org/TR/ttml2/#d3e56767) strongly suggests not permitting entity expansion. We should make it a concrete conformance requirement that XML entities are prohibited.

nigelmegitt:
Resolves #51.

Takes the existing text from IMSC 1.2, changes the SHOULD NOT to a MUST NOT and removes the note about future specification intention to make into a MUST NOT.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/79.html" title="Last updated on Dec 2, 2022, 5:02 PM UTC (dfcc9c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/79/640e525...dfcc9c7.html" title="Last updated on Dec 2, 2022, 5:02 PM UTC (dfcc9c7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -789,10 +789,20 @@ daptm:onScreen
             <li><em>entity references</em> other than to <em>predefined entities</em>.</li>
           </ul>
     
-          <p class="note">The resulting [[xml]] document can contain
+          <div class="note">
+            <p>The resulting [[xml]] document can contain
             <dfn data-cite="xml#dt-charref">character references</dfn>,
             and <dfn data-cite="xml#dt-entref">entity referencess</dfn> to
             <dfn data-cite="xml#sec-predefined-ent">predefined entities</dfn>.</p>
+            <p>The predefined entities are (including the leading ampersand and trailing semicolon):</p>
+            <ul>
+              <li><code>&amp;amp;</code> for an ampersand &amp;</li>
+              <li><code>&amp;apos;</code> for an apostrophe &apos;</li>
+              <li><code>&amp;gt;</code> for a greater than sign &gt;</li>
+              <li><code>&amp;lt;</code> for a less than sign &lt;</li>
+              <li><code>&amp;quot;</code> for a quote symbol &quot;</li>
+            </ul>
+          </div>
       </section>
       <section id="foreign-elements-and-attributes">
         <h3>Foreign Element and Attributes</h3>
--- a/index.html
+++ b/index.html
@@ -778,7 +778,7 @@ daptm:onScreen
       <h2>Constraints</h2>
       <section id="Document Encoding">
         <h3>Document Encoding</h3>
-        <p>A <a>Document Instance</a> SHALL be concretely encoded as a well-formed XML 1.0 [[!xml]] document using the UTF-8
+        <p>A <a>Document Instance</a> MUST be concretely encoded as a well-formed XML 1.0 [[!xml]] document using the UTF-8
           character encoding as specified in [[UNICODE]].</p>
     
           <p>The resulting [[!xml]] document MUST NOT contain any of the following physical structures:</p>
--- a/index.html
+++ b/index.html
@@ -778,8 +778,8 @@ daptm:onScreen
       <h2>Constraints</h2>
       <section id="Document Encoding">
         <h3>Document Encoding</h3>
-        <p>A <a>Document Instance</a> MUST be concretely encoded as a well-formed XML 1.0 [[!xml]] document using the UTF-8
-          character encoding as specified in [[UNICODE]].</p>
+        <p>A <a>Document Instance</a> MUST be serialised as a well-formed XML 1.0 [[!xml]] document
+          encoded using the UTF-8 character encoding as specified in [[UNICODE]].</p>
     
           <p>The resulting [[!xml]] document MUST NOT contain any of the following physical structures:</p>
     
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -607,9 +607,19 @@ <h3>Profile Resolution Semantics</h3>
       <h2>Constraints</h2>
       <section id="Document Encoding">
         <h3>Document Encoding</h3>
-        <p>
-          A Document Instance MUST use UTF-8 character encoding as specified in [[UNICODE]].
-        </p>
+        <p>A <a>Document Instance</a> MUST be concretely encoded as a well-formed XML 1.0 [[!xml]] document using the UTF-8
```
cconcolato, line 610: What does "concretely encoded" means? Can we simply say "... MUST be a well-formed XML 1.0 document ..." ? Also can we get rid of "Document Instance" as discussed in https://github.com/w3c/dapt/issues/57 ?

index.html
```diff
@@ -607,9 +607,19 @@ <h3>Profile Resolution Semantics</h3>
...
+          character encoding as specified in [[UNICODE]].</p>
+    
+          <p>The resulting [[!xml]] document MUST NOT contain any of the following physical structures:</p>
+    
+          <ul>
+            <li><em>entity declarations</em>; and</li>
+    
+            <li><em>entity references</em> other than to <em>predefined entities</em>.</li>
```
cconcolato, line 618: Can we add a link to https://www.w3.org/TR/xml/#sec-predefined-ent ? or make it clear that it means only allowing: `&lt;`, `&gt;`, `&amp;`, `&apos;` and `&quot;`?

index.html
```diff
@@ -607,9 +607,19 @@ <h3>Profile Resolution Semantics</h3>
...
+          </ul>
+    
+          <p class="note">The resulting [[xml]] document can contain <em>character references</em>, and <em>entity references</em> to
```
cconcolato, line 621: Can we add a link to https://www.w3.org/TR/xml/#dt-charref for "character references"? and maybe give an example? Also what's the use case for allowing character references when the encoding is UTF-8?

nigelmegitt: Allowing character references is the default, I think, so I'd look for a use case for disallowing them instead - I can't think of one. In particular, allowing character references can help to unambiguously specify a character, or to know what a character is in case the font being used to review the XML file does not support the character.

nigelmegitt: Links added in ff6d63b.

Thinking more about the idea of prohibiting character references, I think it would be an unwelcome surprise to find that they are not supported, to many. Also, I realise I don't understand why you're linking the use of character references to the UTF-8 encoding, particularly?

index.html
nigelmegitt: Links added in https://github.com/w3c/dapt/commit/ff6d63ba14522e4273796043a1c2596270ae79d9.

index.html
nigelmegitt: Suggest we handle "Document Instance" in #57 in a separate PR.

To me, "concretely encoded" means "turned into a sequence of bits", but I'd be happy to adopt a better term if you have one. I think it's helpful to use the same language as https://www.w3.org/TR/ttml2/#concrete-encoding to avoid confusion, but not if it actually creates confusion!

cconcolato: > Suggest we handle "Document Instance" in https://github.com/w3c/dapt/issues/57 in a separate PR.

Ok

> To me, "concretely encoded" means "turned into a sequence of bits", but I'd be happy to adopt a better term if you have one. I think it's helpful to use the same language as https://www.w3.org/TR/ttml2/#concrete-encoding to avoid confusion, but not if it actually creates confusion!

A quick Google search shows nothing relevant for "XML concrete encoding" or "XML concretely encoded". I would avoid it. Simple saying "A DAPT document MUST be a well-formed XML 1.0 document encoded using UTF-8".

nigelmegitt: I think the distinction we need to allow for is that a document could be either a serialised sequence of bits, in a file system for example, or an in-memory model (DOM). Only the former needs to be encoded in this way. And there's no requirement to serialise a document at all. So we need something a bit different from your suggestion, somehow. Not sure what, yet. Might be worth seeing how others have handled this in other XML flavours.

index.html
cconcolato: Let's say if we were to allow ASCII as a possible encoding, to encode any character you'd have to use character references. With UTF-8 as the encoding you can represent any Unicode code point, thus there is no need for character reference. Not having to support character references would make processors simpler but I won't argue more. We can get feedback from implementers.

index.html
cconcolato: I see TTML2 has even a note about that in https://www.w3.org/TR/ttml2/#concrete-encoding with the exact list. Maybe we should have the same?

nigelmegitt: Fine for me, would be helpful probably.

index.html
cconcolato: I'm not convinced we need the distinction. The whole spec is about a serialized format IMHO. If there was no serialization, no way to exchange documents, there would be no need for the spec. The in-memory model, if needed, follows the serialization, as usual with XML/DOM. I quickly looked at SVG and couldn't find the concept of "document instance" (used in one place, but not relevant).

nigelmegitt: Hmm, will think about this, let's discuss.

index.html
nigelmegitt: Done in 8c2d994.

index.html
nigelmegitt: Done in dfcc9c7.

cconcolato (approved):

Issue #14, opened 2022-06-13
DAPT Data Model: Describe Event and Character


andreastai:
Add text to describe the meaning of the objects/classes event and character.

nigelmegitt:
We think this is completed, please reopen if you don't agree.

andreastai:
Looks good to me.

PR #80, opened 2022-11-30
clarify cardinality of model properties

### LINKED ISSUE #12: DAPT Data Model: Add cardinality to properties

cconcolato:
close #12 

* clarify that "Script Type" is mandatory 
* clarify that "Primary Language" is mandatory
* Remove the definition of "Script Events" (plural) and indicate that a DAPT script has zero or more "Script Event" objects (already defined)
* Remove the definition of "Characters" (plural) and indicate that a DAPT script has zero or more "Character" objects (already defined)
* Clarify in Character that the identifier and name properties are mandatory
* Clarify in Script Event that identifier, begin and end are mandatory, and that character identifiers are optional (on screen text, dialog) and that the description is also optional
* Remove inline reference to issue #12
* Remove inline reference to issue #14 (already closed)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/80.html" title="Last updated on Dec 2, 2022, 4:33 PM UTC (edd7ec2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/80/53c4166...edd7ec2.html" title="Last updated on Dec 2, 2022, 4:33 PM UTC (edd7ec2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -492,7 +492,7 @@ table.coldividers td + td { border-left:1px solid gray; }
             </li>
             <li>a mandatory <dfn>Main Text</dfn> property, which provides the main text content for this <a>Script Event</a> in the form of a <a>Text</a> object</li>
             <li>an optional <dfn>Contextual Text</dfn> property, which provides a representation of the <a>Main Text</a> content in a different language, in the form of a <a>Text</a> object. It can be used to assist in the processing of the main text content.</li>
-            <li>a optional <dfn>Script Event Description</dfn> property, as a human-readable description of the <a>Script Event</a>.
+            <li>an optional <dfn>Script Event Description</dfn> property, as a human-readable description of the <a>Script Event</a>.
               <p class="note">The <a>Script Event Description</a> does not need to be unique, i.e. it does not need to have a different value for each <a>Script Event</a>. 
                 For example a particular value could be re-used to identify in a human-readable way one or more <a>Script Events</a> that are intended to be processed together,
                 e.g. in a batch recording.</p>
--- a/index.html
+++ b/index.html
@@ -397,7 +397,10 @@ table.coldividers td + td { border-left:1px solid gray; }
 
         <p>In <a>Dubbing Scripts</a>, it is necessary to identify each character in the programme. This is done with a <dfn>Character</dfn> object which has the following properties: 
           <ul>
-            <li>a mandatory <dfn data-lt="Character Identifier">Identifier</dfn> which is a unique identifier to indicate when a <a>Character</a> participates in a <a>Script Event</a></li>
+            <li>a mandatory <dfn data-lt="Character Identifier">Identifier</dfn>
+            which is a unique identifier used to reference the character from elsewhere in the document,
+            for example to indicate when a <a>Character</a> participates in a <a>Script Event</a>,
+            or to link a <a>Character Style</a> to its <a>Character</a>.</li>
             <li>a mandatory <dfn data-lt="Character Name">Name</dfn> which is the name of the <a>Character</a> in the programme</li>
             <li>an optional <dfn data-lt="Character Talent Name">Talent Name</dfn>, which is the name of the actor speaking dialogue for this <a>Character</a></li>
             <li>an optional <dfn data-lt="Character Style">Style</dfn> object which can be used to control the style when displaying the <a>Script Event</a> for example during recording by an actor</li>
--- a/index.html
+++ b/index.html
@@ -483,7 +483,7 @@ table.coldividers td + td { border-left:1px solid gray; }
         <p>An <dfn>Script Event</dfn> object represents dialogue, on screen text or audio descriptions to be spoken and has the following properties:
           <ul>
             <li>A mandatory <dfn>Script Event Identifier</dfn> which is unique in the script</li>
-            <li>Mandatory <dfn>Begin</dfn> and <dfn>End</dfn> properties providing the <a>Script Event</a> time interval in the programme timeline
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -385,23 +384,21 @@ <h4>Primary Language</h4>
...
           <section>
             <h4>Script Events</h4>
-            <p>The <dfn>Script Events</dfn> part of the model provides the list of intervals corresponding to dialogue, on screen text, or descriptions.</p>
+            <p>A <a>DAPT Script</a> MAY contain zero or more <a>Script Event</a> objects, each corresponding to dialogue, on screen text, or descriptions for a given time interval.</p>
           </section>
 
           <section>
             <h4>Characters</h4>
-            <p>The <dfn>Characters</dfn> part of the model provides the list of characters referenced in the <a>Script Events</a>.</p>
+            <p>A <a>DAPT Script</a> MAY contain zero or more <a>Character</a> objects, each describing a character referenced in a <a>Script Event</a>.</p>
```
nigelmegitt, line 392: We cannot require that each Character object is referenced in a Script Event, since a workflow might involve importing characters from some other source before any transcription has happened.

```suggestion
            <p>A <a>DAPT Script</a> MAY contain zero or more <a>Character</a> objects, each describing a character that can be referenced by a <a>Script Event</a>.</p>
```

index.html
```diff
@@ -385,23 +384,21 @@ <h4>Primary Language</h4>
...
-          <p class="issue" data-number="14"></p>
       </section>
       <section>
         <h3>Character</h3>
 
         <p>In <a>Dubbing Scripts</a>, it is necessary to identify each character in the programme. This is done with a <dfn>Character</dfn> object which has the following properties: 
           <ul>
-            <li><dfn data-lt="Character Identifier">Identifier</dfn> which is a unique identifier to indicate when a <a>Character</a> participates in a <a>Script Event</a></li>
-            <li><dfn data-lt="Character Name">Name</dfn> which is the name of the <a>Character</a> in the programme</li>
+            <li>a mandatory <dfn data-lt="Character Identifier">Identifier</dfn> which is a unique identifier to indicate when a <a>Character</a> participates in a <a>Script Event</a></li>
```
nigelmegitt, line 400: Noticing that the phrasing here should be adjusted (could be a separate issue but may as well deal with it here for expediency): the identifier does not indicate when a Character participates in a Script Event; rather, the identifier can be used to reference a Character.

```suggestion
            <li>a mandatory <dfn data-lt="Character Identifier">Identifier</dfn>
            which is a unique identifier used to reference the character from elsewhere in the document,
            for example to indicate when a <a>Character</a> participates in a <a>Script Event</a>,
            or to link a <a>Character Style</a> to its <a>Character</a>.</li>
```

index.html
```diff
@@ -485,17 +482,17 @@ <h3>Character</h3>
         <h3>Script Event</h3>
         <p>An <dfn>Script Event</dfn> object represents dialogue, on screen text or audio descriptions to be spoken and has the following properties:
           <ul>
-            <li>An <dfn>Script Event Identifier</dfn> which is unique in the script</li>
-            <li><dfn>Begin</dfn> and <dfn>End</dfn> properties providing the <a>Script Event</a> time interval in the programme timeline
+            <li>A mandatory <dfn>Script Event Identifier</dfn> which is unique in the script</li>
+            <li>Mandatory <dfn>Begin</dfn> and <dfn>End</dfn> properties providing the <a>Script Event</a> time interval in the programme timeline
```
nigelmegitt, line 486: Not quite clear that both properties are mandatory, somehow.

```suggestion
            <li>A mandatory <dfn>Begin</dfn> property and a mandatory <dfn>End</dfn> property that together define the <a>Script Event</a>'s time interval in the programme timeline
```

index.html
```diff
@@ -485,17 +482,17 @@ <h3>Character</h3>
...
+            <li>Zero or more <a>Character Identifiers</a> indicating the <a>Characters</a> involved in this <a>Script Event</a>.
               <p class="note">While typically, a <a>Script Event</a> corresponds to one single <a>Character</a>, there are cases where multiple <a>characters</a> can be associated with a <a>Script Event</a>. This is when all <a>Characters</a> speak the same text at the same time.</p>
             </li>
             <li>a mandatory <dfn>Main Text</dfn> property, which provides the main text content for this <a>Script Event</a> in the form of a <a>Text</a> object</li>
             <li>an optional <dfn>Contextual Text</dfn> property, which provides a representation of the <a>Main Text</a> content in a different language, in the form of a <a>Text</a> object. It can be used to assist in the processing of the main text content.</li>
-            <li>a <dfn>Script Event Description</dfn> property, as a human-readable description of the <a>Script Event</a>.
+            <li>a optional <dfn>Script Event Description</dfn> property, as a human-readable description of the <a>Script Event</a>.
```
nigelmegitt, line 495: Grammar nit:

```suggestion
            <li>an optional <dfn>Script Event Description</dfn> property, as a human-readable description of the <a>Script Event</a>.
```

nigelmegitt (changes requested):
A few points with suggestions, hopefully easily resolved.

index.html
cconcolato: In a finished document, I think it would be simpler if only those characters that are used are present in a document. But maybe that's imposing too much. Should be a 'should'.

index.html
cconcolato: But the Character should be referenced only when it participates in the event, no?

index.html
cconcolato: Ok to use your suggestion and revisit when discussing https://github.com/w3c/dapt/issues/84

index.html
cconcolato: We need to make sure that the requirement is expressed in the TTML section.

nigelmegitt (approved):

PR #81, opened 2022-11-30
define a type attribute in DAPT metadata namespace to annotate script events

### LINKED ISSUE #35: [ED Issue] How to add event type information?
From ED Issue under Document Structure -> Event -> `ttm:desc`:
>How to add [Event type](https://w3c.github.io/dapt/#dfn-event-type) information? ttm:role? What if already used by the onscreen annotation?

cconcolato:
close #35 

I simply used "type" instead of "descType" as discussed in the issue because we already use `ttm:desc` for something else and I wanted to avoid the confusion. I left an editor's note to discuss different options.


cconcolato:
I should add the attribute to the existing examples.

nigelmegitt (changes requested):
I'm confused here. #35 suggests that we want to define an attribute for the `ttm:desc` element, but we have a number of problems here, which I think we need to take a run at together:

* It isn't clear if we're defining an attribute that should be set on the Script Event `div` directly, to define the type of the script event, or an attribute that should be set on a `ttm:desc` element to qualify what sort of description the element contains.
* Script Event Description is defined within the On Screen section.
* Script Event Type is also defined within the On Screen section.

I suggest:

* We create separate sections for each, outside On Screen.
* We add an enum / registry property to define the type of a Script Event
* We add an enum / registry property to define the type of a Script Event Description.

Would that work?

cconcolato:
We agree to change the PR as follows:
* define a descType on the desc element, as string
* rename type as eventType
* add a note to make a registry for these 2 elements

index.html
```diff
@@ -744,15 +740,55 @@ <h4>On Screen</h4>
...
+            e.g. in a batch recording.</p>
+        <pre class="example"
+          data-include="examples/event-desc.xml"
+          data-include-format="text">
+        </pre>
+          <p>The <code>ttm:desc</code> element MAY have a <code>daptm:descType</code> attribute specified to indicate the type of description. The <code>daptm:descType</code> attribute is defined below. Its possible values are as indicated in the registry at YYY.</p>
+          <p class="ednote">Registry to be defined.</p>
+          <div class="exampleInner">
+            <pre class="nohighlight">
+daptm:eventType = string
```
nigelmegitt, line 762:
```suggestion
daptm:descType = string
```

examples/event-desc-descType.xml
```diff
@@ -0,0 +1,26 @@
+...
+  <body>
+    <div begin="10s" end="13s">
+      <metadata>
+        <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
```
nigelmegitt, line 5: Probably worth putting two different `ttm:desc`s in with different types to make it clear why we're bothing, for example maybe something like:

```suggestion
        <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
        <ttm:desc daptm:descType="plotSignificance">High</ttm:desc>
```

At https://w3c.github.io/adpt/#bbc-example-section the two values used are for "pronunciation note" and "shot description" - if they sound like better proposals for the example I'm happy to use those instead.

Also they're both `ttm:desc` elements that are children of the `<p>` element, i.e. no need to make them children of a `<metadata>` element - this is clearly allowed in TTML2, by the way, so may be worth doing as well, as a simplification?

examples/event-desc-descType.xml
```diff
@@ -0,0 +1,26 @@
...
+      <p daptm:langSrc="original">
+        <span>A woman climbs into a small sailing boat.</span>
+      </p>
+      <p xml:lang="fr" daptm:langSrc="translation">
+        <span>Une femme monte à bord d'un petit bateau à voile.</span>
+      </p>
+    </div>
+    <div begin="18s" end="20s">
+      <metadata>
+        <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
```
nigelmegitt, line 16:
```suggestion
        <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
        <ttm:desc daptm:descType="plotSignificance">Low</ttm:desc>
```

index.html
```diff
@@ -744,15 +740,55 @@ <h4>On Screen</h4>
...
+          <h4>Script Event Description</h4>
+          <p>The <dfn>Script Event Description</dfn> property is an annotation providing a human-readable description of a <a>Script Event</a>.
+          <p>The <a>Script Event Description</a> property is represented in TTML as <code>ttm:desc</code> element added to a <code>metadata</code> element at the <code>div</code> level.</p>
+          <p class="note">The <a>Script Event Description</a> does not need to be unique, i.e. it does not need to have a different value for each <a>Script Event</a>. 
+            For example a particular value could be re-used to identify in a human-readable way one or more <a>Script Events</a> that are intended to be processed together,
+            e.g. in a batch recording.</p>
+        <pre class="example"
+          data-include="examples/event-desc.xml"
+          data-include-format="text">
+        </pre>
```
nigelmegitt, line 757: This example is so similar to the following one with `descType` that I don't think we need both. Suggest removing this and the file, unless I've missed something.

nigelmegitt (changes requested):
Almost there...

index.html
cconcolato: I expect the example with no descType to be the most frequent case. I'd rather keep it.

nigelmegitt: Can we simplify the non-descType example then, to include just a `<div>`, and maybe take out the `<metadata>` element?

cconcolato: Oh, I see... `ttm:desc` is in the `Metadata.class` and can be put directly under the `div` without needing the `metadata` element. I can fix that.
Regarding simplification of the non-descType, the idea of the example is to show that the same description can be used for 2 consecutive events. I'm not sure it makes sense to remove the content.
I think we're struggling here because the use of `descType` is not clear. Maybe describing more the second example to say something like: "This example is equivalent to the previous one except that the actual type of event description is given when disambiguation is needed".

nigelmegitt: Sure, two different examples are fine, but not so similar. How about something like:

ttm:desc non-uniqueness:

```xml
<div [attributes] daptm:onScreen="OFF">
  <ttm:desc>Voiceover</ttm:desc>
  <p daptm:langSrc="original">Will she remember how to sail, after all these years?</p>
</div>

<div [attributes] daptm:onScreen="OFF">
  <ttm:desc>Voiceover</ttm:desc>
  <p daptm:langSrc="original">Of course!</p>
</div>
```

ttm:desc with descType:

```xml
<div [attributes]>
  <ttm:desc daptm:descType="pronunciationNote">[oːnʲ]</ttm:desc>
  <p daptm:langSrc="original">Eóin looks around at the other assembly members.</p>
</div>
```

cconcolato:
There are many annotations that one could attach to an event:
* language 
* languageSource (enumeration)
* scene information (free form text)
* plot significance (enumeration)
* on/off screen (enumeration)
* type of event (on screen, dialog, description, ...)
* emotion (probably enumeration)
* ...

We have several tools at our disposal:
* a new attribute in the `daptm` namespace
* a `ttm:desc` element with `daptm:descType` attribute. 
* a `metadata` element
* attributes/elements in foreign namespaces
* ...

For the values in the elements or in the attributes, we have the choice between:
* single value or multiple values (space separated, comma separated, ...)
* each value being: a general string, from an enumeration in some registry, or from an enumeration in this spec

I wonder if we shouldn't have a section "Script Event Annotations", with an introduction describing the general design principles  and then sub sections for each annotation.

cconcolato:
Done, issue created https://github.com/w3c/dapt/issues/100
Can we merge this one as is or do you still want me to tweak the examples?

nigelmegitt:
I think it's worth tweaking the examples before merging, please.

cconcolato:
@nigelmegitt I think I addressed your comments. The descType section now has 3 examples: one with simply `ttm:desc` and no `daptm:descType`; one with both; and one with 2 `daptm:descType`.

index.html
```diff
@@ -781,6 +781,11 @@ <h4>Script Event Description</h4>
             data-include="examples/event-desc-descType.xml"
             data-include-format="text">
           </pre>
+          <p>Multiple <code>&lt;ttm:desc&gt;</code> elements MAY be present with different values <code>daptm:descType</code>, as in the following example.</p>
```
nigelmegitt, line 784: Minor nit:

```suggestion
          <p>Multiple <code>&lt;ttm:desc&gt;</code> elements MAY be present with different values of <code>daptm:descType</code>, as in the following example.</p>
```

index.html
```diff
@@ -762,15 +762,15 @@ <h4>On Screen</h4>
         <section>
           <h4>Script Event Description</h4>
           <p>The <dfn>Script Event Description</dfn> property is an annotation providing a human-readable description of a <a>Script Event</a>.
-          <p>The <a>Script Event Description</a> property is represented in TTML as <code>ttm:desc</code> element added to a <code>metadata</code> element at the <code>div</code> level.</p>
+          <p>The <a>Script Event Description</a> property is represented in TTML as <code>&lt;ttm:desc&gt;</code> element at the <code>&lt;div&gt;</code> level.</p>
```
nigelmegitt, line 765: Minor nit:

```suggestion
          <p>The <a>Script Event Description</a> property is represented in TTML as a <code>&lt;ttm:desc&gt;</code> element at the <code>&lt;div&gt;</code> level.</p>
```

nigelmegitt (approved):
Thanks looks good, barring a couple of tiny grammar nits.

PR #85, opened 2022-12-02
replace ttm:role by daptm:scriptType

### LINKED ISSUE #24: ttm:role values - we should be able to define new ones in the registry
TTML2 uses a Registry for ttm:role values - we should be able to define new values in that registry without the `X-` prefix, in DAPT, since it's a W3C TTWG spec.

### LINKED ISSUE #32: [ED issue] should we define ttm:role attribute values in this specification?
Was ED Issue text:
>Should these values be defined in this specification? Or should we reuse/extend existing specifications?

See also #24

### LINKED ISSUE #33: [ED issue] We need to define values of `ttm:role` for Audio Description script type(s).
From ED issue text under Document Structure for Scripts:
>We need to define values for Audio Description script type(s).

### LINKED ISSUE #54: Rather than using ttm:role for script type, define new attribute
Rather than defining new values for `ttm:role` for script type, I suggest we create a new attribute in the `daptm:` namespace called `scriptType` and make its values a Registry track list.

We can then include this in the `tt` element so that minimal (SAX) parsing is required before discovering the script type.

cconcolato:
close #24 
close #32 
close #33
close #54

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/85.html" title="Last updated on Dec 2, 2022, 9:54 PM UTC (c6a153c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/85/a2fec52...c6a153c.html" title="Last updated on Dec 2, 2022, 9:54 PM UTC (c6a153c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -346,21 +346,23 @@ table.coldividers td + td { border-left:1px solid gray; }
           <section>
             <h4>Script Type</h4>
             <p>The <dfn>Script Type</dfn> property is a mandatory property of a <a>DAPT Script</a> which describes the type of documents used in Dubbing and Audio Description workflows, among the following: <a>Original Language Dialogue List</a>, <a>Translated Dialogue List</a>, <a>Pre-recording Dub Script</a>, <a>As-recorded Dub Script</a>, and <a>Audio Description Script</a>. <a>Original Language Dialogue List</a>, <a>Translated Dialogue List</a>, <a>Pre-recording Dub Script</a>, <a>As-recorded Dub Script</a> are referred to as <dfn>Dubbing Scripts</dfn>.</p>
-            <p>To represent this property, the following TTML structure and constraints apply:
-              <ul>
-                <li>The <code>head</code> element MUST be present, as the first child of the <code>tt</code> element</li>
-                <li>A single <code>metadata</code> element MUST be present in the <code>head</code> element with the <code>ttm:role</code> attribute set to one of the values defined below.
-                </li>
-              </ul>
+            <p>To represent this property, the following TTML attribute MUST be present on the <code>tt</code> element:</p>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -346,21 +346,23 @@ <h3>DAPT Script</h3>
...
-                <li>The <code>head</code> element MUST be present, as the first child of the <code>tt</code> element</li>
-                <li>A single <code>metadata</code> element MUST be present in the <code>head</code> element with the <code>ttm:role</code> attribute set to one of the values defined below.
-                </li>
-              </ul>
+            <p>To represent this property, the following TTML attribute MUST be present on the <code>tt</code> element:</p>
+            <div class="exampleInner">
+              <pre class="nohighlight">
+  daptm:scriptType
+    : "DUBBING_ORIGINAL_DIALOGUE_LIST"
+    | "DUBBING_TRANSLATED_DIALOGUE_LIST"
```
nigelmegitt, line 354: Looking at these values, the suffix `_LIST` looks out of place. And since we're now being explicit about it possibly containing on screen text as well as dialogue, the `DIALOGUE` bit looks odd too. Happy to open as a separate issue, but I'm leaning towards just `ORIGINAL_TRANSCRIPT` and `TRANSLATED_TRANSCRIPT`, partly because I think they are clearer and also because I suspect they are likely to be more generally useful beyond just dubbing workflows.

I'd keep `DUBBING_PRE_RECORDED` and `DUBBING_AS_RECORDED` as they are.

nigelmegitt (approved):
Looks good to me. Can open an issue for my comment, or deal with it now, either way happy to merge.

PR #86, opened 2022-12-06
Add UML class diagram to help explain the data model

### LINKED ISSUE #55: [ED issue] Consider adding a diagram of the DAPT Script model
We could use a diagram that shows the nested objects (script, event, character, ...) and the properties of each.

nigelmegitt:
Closes #55.

PlantUML source and generated SVG, including fragment links into the document, plus data include to bring it into the Data Model section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/86.html" title="Last updated on Dec 12, 2022, 10:42 AM UTC (48dc849)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/86/4a1bd2e...48dc849.html" title="Last updated on Dec 12, 2022, 10:42 AM UTC (48dc849)">Diff</a>


images/sources/class-diagram.puml
```diff
@@ -0,0 +1,53 @@
...
+}
+
+"DAPT Script" "0" *-- "many" "Script Event" : contains
+"DAPT Script" "0" *-- "many" "Character" : contains
+"DAPT Script" "0" *-- "many" "Style" : contains
+"Character" "0" .. "many" "Style" : Character Style
+"Script Event" "0" .. "many" "Style" : Character Style
+"Text" "0" .. "many" "Style" : Character Style
+"Text" "0" .. "many" "Style" : Other Style
+"Script Event" "0" *-- "many" "Text" : contains
```
cconcolato, line 48: Can we replace many by 2? A Script Event contains at most 2 Text objects.

nigelmegitt: I don't think so, as things stand. For example a Script Event can have Original language, Pivot translation and Target language translation which would have to be 3 Text objects. If there's a need to constrain the contents per script type, that would be an additional constraint, but in general, it's 0 to many as we've written the spec at the moment.

nigelmegitt: By the way, all the cardinalities made no sense and have now been fixed to have ranges on both ends of the arrow where appropriate.

PR #91, opened 2022-12-07
Inline definitions and remove the Definitions section

### LINKED ISSUE #58: [ED issue] Consider removing the definition section
Terms could be defined the first time they are used and linked when they are used subsequently. We don't really need a definition section. Also, we could let respec generate the index section https://respec.org/docs/#id-index

cconcolato:
close #58

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/91.html" title="Last updated on Dec 7, 2022, 8:12 PM UTC (d8b40e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/91/4dfeef2...d8b40e9.html" title="Last updated on Dec 7, 2022, 8:12 PM UTC (d8b40e9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -236,9 +236,6 @@ table.coldividers td + td { border-left:1px solid gray; }
         The following terms are used in this proposal:
       </p>
       <p>
-        <dfn data-cite="ttml2#style-value-lwsp">Linear White-Space</dfn> See Section 2.3 at [[TTML2]].
-      </p>
-      <p>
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
This has moved a bunch of definitions from normative sections to informative ones. We probably shouldn't do that in general, but at the moment all the references to those informative definitions are themselves informative, so it's okay, just something we need to watch out for...

1 suggestion, but it's okay without, better with.

PR #101, opened 2022-12-12
Improve introduction

nigelmegitt:
Closes #28.

Also adds the flag for Respec to indicate that this is intended to be an updateable Rec.

* Introduce transcripts and DAPT in general, describing what's in common across dubbing and AD.
* Make Dubbing and Audio description each a sub-section of the first introduction section.
* Some editorial tweaks to the pre-existing text as well as moving it.
* Describe the AD production steps better.
* Fix a reference to the Namespaces section so that it comes out with the right TOC number.
* Fix a couple of Respec definition not-being-referenced errors.

Incorporates changes previously in #102:

Closes #87.

* Change `DUBBING_ORIGINAL_DIALOGUE_LIST` to `ORIGINAL_TRANSCRIPT` and change the name "Original Language Dialogue List" to "Original Language Transcript"
* Change `DUBBING_TRANSLATED_DIALOGUE_LIST` to `TRANSLATED_TRANSCRIPT` and change the name "Translated Dialogue List" to "Translated Transcript"
* Split out the list of script types that are Dubbing Scripts into a separate paragraph for readability.
* Propagate changes throughout the examples
* Add an Ed Note before Example 12 - looks like it's orphaned and needs to move higher up, possibly.
* Editorial improvements to the text about the Audio Description script type.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/101.html" title="Last updated on Jan 19, 2023, 9:52 AM UTC (8346073)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/101/545900c...8346073.html" title="Last updated on Jan 19, 2023, 9:52 AM UTC (8346073)">Diff</a>

PR #102, opened 2022-12-16
Rename DUBBING_ORIGINAL_DIALOGUE_LIST and DUBBING_TRANSLATED_DIALOGUE_LIST

### LINKED ISSUE #87: Rename DUBBING_ORIGINAL_DIALOGUE_LIST and DUBBING_TRANSLATED_DIALOGUE_LIST
Looking at the values of Script Type, the suffix `_LIST` looks out of place in the two in which it occurs.

>        Looking at these values, the suffix `_LIST` looks out of place. And since we're now being explicit about it possibly containing on screen text as well as dialogue, the `DIALOGUE` bit looks odd too. Happy to open as a separate issue, but I'm leaning towards just `ORIGINAL_TRANSCRIPT` and `TRANSLATED_TRANSCRIPT`, partly because I think they are clearer and also because I suspect they are likely to be more generally useful beyond just dubbing workflows.

I'd keep `DUBBING_PRE_RECORDED` and `DUBBING_AS_RECORDED` as they are.

_Originally posted by @nigelmegitt in https://github.com/w3c/dapt/pull/85#discussion_r1040858711_

nigelmegitt:
Closes #87.

* Change `DUBBING_ORIGINAL_DIALOGUE_LIST` to `ORIGINAL_TRANSCRIPT` and change the name "Original Language Dialogue List" to "Original Language Transcript"
* Change `DUBBING_TRANSLATED_DIALOGUE_LIST` to `TRANSLATED_TRANSCRIPT` and change the name "Translated Dialogue List" to "Translated Transcript"
* Split out the list of script types that are Dubbing Scripts into a separate paragraph for readability.
* Propagate changes throughout the examples
* Add an Ed Note before Example 12 - looks like it's orphaned and needs to move higher up, possibly.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/102.html" title="Last updated on Dec 16, 2022, 1:57 PM UTC (e8e771c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/102/b2a93ab...e8e771c.html" title="Last updated on Dec 16, 2022, 1:57 PM UTC (e8e771c)">Diff</a>

### DIFF
--- a/examples/intro-original-language-with-dub-language.xml
+++ b/examples/intro-original-language-with-dub-language.xml
@@ -3,7 +3,7 @@
     xmlns:daptm="http://www.w3.org/ns/ttml/profile/dapt#metadata"
     xml:lang="en"
     ttp:contentProfiles="http://www.w3.org/ns/ttml/profile/dapt"
-    daptm:scriptType="DUBBING_TRANSLATED_DIALOGUE_LIST">
+    daptm:scriptType="TRANSLATED_TRANSCRIPT">
   <head>
     <metadata>
       <ttm:agent type="character" xml:id="character_1">
--- a/examples/intro-original-language.xml
+++ b/examples/intro-original-language.xml
@@ -3,7 +3,7 @@
     xmlns:daptm="http://www.w3.org/ns/ttml/profile/dapt#metadata"
     xml:lang="fr" 
     ttp:contentProfiles="http://www.w3.org/ns/ttml/profile/dapt"
-    daptm:scriptType="DUBBING_ORIGINAL_DIALOGUE_LIST">
+    daptm:scriptType="ORIGINAL_TRANSCRIPT">
   <head>
     <metadata>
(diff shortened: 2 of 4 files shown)


cconcolato (commented):
Can we simply use "Script" instead of "Transcript"? At least for consistency with the other values.

nigelmegitt:
> Can we simply use "Script" instead of "Transcript"? At least for consistency with the other values.

I'd rather not - see also #101 where I try to make the distinction clearer and more useful.

nigelmegitt:
This will have conflicts with #101 but is broadly aligned with the direction of change in that pull request too.

nigelmegitt:
I've merged this into #101, so closing this. Makes sense to review them together in the context of improving the consistency of how the spec uses the terminology "script" and "transcript".

Issue #71, opened 2022-09-27
How would extended AD be supported?


nigelmegitt:
@nigelmegitt - audio in the case of dub/vo, but for AD, the video is the important association!

Not sure if you have considered 'Extended AD' - where the AD presentation of the original media may cause the original media to pause at certain points for the AD to complete before playing again.
https://www.w3.org/TR/WCAG20-TECHS/G8.html#:~:text=Extended%20audio%20description%20temporarily%20pauses,are%20insufficient%20for%20adequate%20description.

If not, then it's probably a DAPT wishlist item, as I'm not aware of good examples of preparing such a thing at the moment.
It's not something we've implemented yet - A method of representing it is evading me so far (although I've not thought too hard yet).

_Originally posted by @btsimonh in https://github.com/w3c/dapt/issues/45#issuecomment-1250026733_

nigelmegitt:
My thoughts on this:

The extension that Extended AD gives relative to "ordinary" AD is that it allows the audio description resource to be longer than the period within the related media that is set aside for playback of that resource.

In other words, within, say, a 2s period of media there is a 5s duration description.

For Text to Speech applications, it is very often not known at authoring time exactly how long the speech will last.

For pre-recorded audio, the duration of each description is known exactly, after it has been recorded.

The current semantics of TTML2 playback of audio is that when the end time of the `<span>` (for TTS) or `<audio>` element is reached, playback is stopped. This is independent of how far through playback of its audio resource it has reached: if it has reached the end, fine; if it has not, it will be truncated.

From a player perspective, it is not obvious whether any truncation is intended or accidental, so we cannot just set a playback flag saying "until audio playback is complete, pause media timeline" because for AD files authored that _rely_ on the truncation behaviour will be undesirable. This could be discouraged of course.

It probably makes sense then to add some syntax whose semantic is:
1. "the actual duration of this audio element needs to be extended to `[duration]`" (comes into play when the "opportunity" time, i.e. the active duration of the `<audio>` element is less than `duration`) and
2. (possibly) "if you have to pause the media, do it at time `M`".

Then it would be a player flag that says "honour extended durations" that would do so by pausing for `max(duration - opportunity, 0)` at either: 1. the defined pause time `M` or 2. some implementation-defined time.

For text to speech playback, the pause time would probably need to be at the end of the "opportunity" window, while the player waits until the text to speech system fires a "completed utterance" event.

[...]

PR #118, opened 2023-03-17
Add audio and audio mixing

### LINKED ISSUE #105: Add audio mixing into data model and TTML representation
We say surprisingly little at the moment about how to define audio mixing and link in audio resources. It's technically there via the profile definition, but it needs to be added to the data model and the TTML representation so it's clear how to use it.

nigelmegitt:
Closes #105 and #71.

* Add Audio, Synthesized Audio, Audio Recording and Mixing Instruction to the data model
* Add informative section explaining audio mixing, with figures
* Add extensions and apply constraints to content and processor profiles to support the required audio features
* Reference "question" issues about the ways to reference or embed audio resources
* Create the concept of "shared properties" and a subset of those that are "timing properties", so that Begin, End and Duration can be defined once and reused in all the classes where they apply
* Fix some linting issues
* Explain how extended descriptions can be implemented but are not specified
* Define handling of multiple `<source>` element children of `<audio>`


cconcolato (commented):
My overall comment on the PR is not about the editing (I have minor comments/questions) but more about the complexity of the solution. We are leaving too many options: src vs source, fragment identifiers, external files vs embedded, base64 vs others, audio codecs, multiple sources .... I would prefer if we went for a simple solution first and then add more features.

In terms of editing, some comments:
- can we integrate section 5 inside 4.10 or move it as an Annex?
- the spec is starting to get big (47 pages!) The diagrams are taking a lot of space (and some of the open issues). I'm wondering if we should move the large diagrams to annexes.

nigelmegitt:
> We are leaving too many options: src vs source, fragment identifiers, external files vs embedded, base64 vs others, audio codecs, multiple sources .... I would prefer if we went for a simple solution first and then add more features.

I agree - the difficulty I have is in knowing which of the options to retain and which to omit. I want industry feedback so I have raised [question issues](https://github.com/w3c/dapt/issues?q=is%3Aissue+is%3Aopen+label%3Aquestion) about them. Seems like the best idea to me is to see what people want and need and remove everything else.

nigelmegitt:
> * the spec is starting to get big (47 pages!) The diagrams are taking a lot of space (and some of the open issues). I'm wondering if we should move the large diagrams to annexes.

In my view, that would not be an improvement. By closing open questions issues we will save some space, for sure. I'm not even sure that the length is a problem _per se_ - I think we're targeting readability and conceptual explanations over length considerations.

I just did a print to pdf preview and on my machine at A4 size it comes to 44 pages, of which 23 are the main body - data model, constraints and conformance, and they start on page 8. Everything from page 32 onwards is appendix.

cconcolato (approved):

PR #120, opened 2023-03-17
Update synchronization section

### LINKED ISSUE #45: [ED issue] Consider synchronization and relationship to audio/video programme
The DAPT requirements document illustrates that workflows start with an audio/video asset. We should identify if anything needs to be said about any synchronization issues (e.g. as done in IMSC) and if the presence of a "Related Media Object" is always assumed.

nigelmegitt:
Closes #45.

* Relate the document instance to a related audio object as well as a related video object.
* Synchronization basis is the audio sample rather than the video frame.
* If using for authoring IMSC, IMSC synchronisation provisions apply.
* If real time playback cannot achieve audio sample accuracy, it SHOULD be within 10ms.
* Update the wording of the note in the synchronisation section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/120.html" title="Last updated on Mar 28, 2023, 4:58 PM UTC (90ffa00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/120/b5376b4...90ffa00.html" title="Last updated on Mar 28, 2023, 4:58 PM UTC (90ffa00)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -32,10 +32,16 @@
       localBiblio: {
           "BBC-WHP051": {
             title: "BBC R&D White Paper WHP 051. Audio Description: what it is and how it works",
-            publisher: " N.E. Tanton, T. Ware and M. Armstrong",
+            publisher: "N.E. Tanton, T. Ware and M. Armstrong",
             status: "October 2002 (revised July 2004)",
             href: "http://www.bbc.co.uk/rd/publications/whitepaper051",
           },
+          "EBU-R37": {
+            title: "EBU Recommendation R37-2007. The relative timing of the sound and vision components of a television signal",
+            publisher: "EBU/UER",
+            status: "February 2007",
+            href: "https://tech.ebu.ch/publications/r037"
+          }
       },
     };
     </script>
@@ -473,7 +479,7 @@ table.coldividers td + td { border-left:1px solid gray; }
                 <dfn>Audio Description Script</dfn>:
                 <p>When the <code>daptm:scriptType</code> value is <code>AUDIO_DESCRIPTION</code>,
                   the document represents a <a>transcript</a> of the important visual contents of the
-                  <a>Related Video Object</a> written to be suitable for use as a <a>script</a>
+                  video written to be suitable for use as a <a>script</a>
                   for generating an audio representation.
                   It contains text, times, and optional links to audio and mixing instructions
                   for the purpose of producing an audio description audio track.</p>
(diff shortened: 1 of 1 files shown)


cconcolato (changes requested):
Regarding section "5.4 Related Video Object", I'm not sure why we need to reuse these formal terms (RMV, RMA, RMO) from TTML2 and why not simply talk about the associated audio/video content as is done in the rest of the spec. Also, in the detailed definitions, it's not clear why we need to be so precise, and say that an image is a rectangular array of pixels or that audio in multiple channels are "co-synchronous". The use of MAY in 5.4 is also ambiguous. There is always an associated audio/video programme. Maybe the video part is missing when doing some dubbing tasks. Maybe the audio part is missing when doing some audio description tasks, but these are edge cases.
Outside of the clauses "Related Video Object" and "Synchronization", these terms are used in:
- "4.1.1 Script Type" in the definition of "Audio Description Script". 
- "4.3 Script Event" in a note on the absence of end attribute
- "4.5 Text Language Source" to indicate that the source of the language. Again not essential.
- "5.7.1 ttp:timeBase" to indicate that time 0 in the script corresponds to time 0 in the RMO.
None of these usages seem essential. 

I suggest deleting the section and removing the use of the terms, or simply replacing by "associated programme", "audio/video content", or alike, as already done throughout the text.

Regarding section "5.5 synchronization", my understanding of the matching section in IMSC1.2 https://www.w3.org/TR/ttml-imsc1.2/#synchronization is that it is there to fine-tune the case where the refresh rate of the screen does not allow subtitles to start at arbitrary times and in that case IMSC requires snapping the caption rendering at the previous frame start. I don't think we need that in DAPT (I'm not sure it's implemented for IMSC anyway and it'd be hard to verify). It's a rendering consideration that belongs to IMSC. The need for the additional snapping to audio samples is not clear to me either. Regarding audio mixing consideration, the use of 10ms seems arbitrary and it relates more to a quality of implementation than a specification part, like using or not anti-aliasing when rendering subtitles. I'm not sure we need that.

I suggest deleting the section.

nigelmegitt:
> The use of MAY in 5.4 is also ambiguous. There is always an associated audio/video programme. Maybe the video part is missing when doing some dubbing tasks. Maybe the audio part is missing when doing some audio description tasks, but these are edge cases.

That's exactly why the conformance language is MAY - I think you've answered your own point there!

On looking again at those sections I see that there are some editorial errors in there, and I will simplify it, at the risk of making pointless changes, at least so it's a bit clearer what we may delete.

I agree that the IMSC fine tuning does not apply here, but I do think that synchronization is a really important so we need to say _something_ about it, rather than deleting the section altogether.

I also agree that we don't especially need a formal definition of related audio and related video objects, except in the sense that we want to be clear about the relationship in TTML terms. Will look at how to edit those to retain that sense.

nigelmegitt:
Significant changes made in response to your comments @cconcolato but I didn't do exactly as you suggested - please re-review.

nigelmegitt:
editorial changes to make:

errant ; in 5.4
specific samples of the audio -> specific audio samples
exact sample time mapping -> rephrase to be clearer e.g. if "accurate synchronization to the audio sample"
60s -> 60ms

PR #127, opened 2023-04-06
Prohibit frame components in times

### LINKED ISSUE #123: Consider restricting time expressions
Given that the time expression syntax for `#time-clock-with-frames` can be misinterpreted as a SMPTE time code, which are not supported, we should consider not permitting `#time-clock-with-frames` and only support `hh:mm:ss.ms` or `f`.

nigelmegitt:
Closes #123.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/127.html" title="Last updated on Apr 10, 2023, 12:58 PM UTC (14589e1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/127/daf097f...14589e1.html" title="Last updated on Apr 10, 2023, 12:58 PM UTC (14589e1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1569,9 +1569,9 @@ daptm:eventType : string
           <h4>Time expressions</h4>
           <p>All time expressions within a document SHOULD use the same syntax,
             either <code>clock-time</code> or <code>offset-time</code>
-            as defined in [[ttml2]].</p>
+            as defined in [[ttml2]], with <a href="#features-and-extensions-section">DAPT constraints</a> applied.</p>
           <div class="example note">
-            <p>A <code>clock-time</code> has one of the forms:</p>
+            <p>A DAPT <code>clock-time</code> has one of the forms:</p>
             <ul>
               <li><code>hh:mm:ss.sss</code></li>
               <li><code>hh:mm:ss</code></li>
--- a/index.html
+++ b/index.html
@@ -1554,32 +1554,11 @@ daptm:eventType : string
             the computed begin time of its parent element,
             or for the <code>&lt;body&gt;</code> element, to time zero.</p>
         </section> <!-- timecontainer-section -->
-        <section id="framerate-section">
-          <h4><code>ttp:frameRate</code></h4>
-          <p>If the document contains any time expression that uses the <code>f</code> metric,
-            or any time expression that contains a frames component,
-            the <code>ttp:frameRate</code> attribute MUST be present on the <code>&lt;tt&gt;</code> element.</p>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -1554,32 +1554,11 @@ <h4><code>timeContainer</code></h4>
             the computed begin time of its parent element,
             or for the <code>&lt;body&gt;</code> element, to time zero.</p>
         </section> <!-- timecontainer-section -->
-        <section id="framerate-section">
-          <h4><code>ttp:frameRate</code></h4>
-          <p>If the document contains any time expression that uses the <code>f</code> metric,
-            or any time expression that contains a frames component,
-            the <code>ttp:frameRate</code> attribute MUST be present on the <code>&lt;tt&gt;</code> element.</p>
-        </section> <!-- framerate-section -->
```
cconcolato, line 1562: I'm not sure we want to remove `f`. The only thing we want to remove is the syntax that looks like SMPTE time code.

index.html
```diff
@@ -2177,13 +2156,13 @@ <h4>Disposition of Features and Extensions</h4>
...
-            <td>See [[[#framerate-section]]] and [[[#time-clock-with-frames-section]]].</td>
+            <td><span class="prohibited label">prohibited</span></td>
+            <td></td>
           </tr>
           <tr>
             <td><code>#time-offset-with-frames</code></td>
-            <td><span class="permitted label">permitted</span></td>
-            <td>See [[[#framerate-section]]].</td>
+            <td><span class="prohibited label">prohibited</span></td>
+            <td></td>
```
cconcolato, line 2165: This one should still be allowed

index.html
nigelmegitt: Seems like unnecessary complexity to support frame counts if no other frame based expressions are permitted. Why not use ticks if an arbitrary rate time counting mechanism is needed?

I've realised I am not sure if subframes have any bearing on frame count time expressions.

nigelmegitt: I see from the issue that the proposal was to retain `f` units so I should have checked more carefully. Will fix when I can.

index.html
```diff
@@ -1590,19 +1575,23 @@ <h4>Time expressions</h4>
             <ul>
```
cconcolato, line 1575: Maybe we should say "A clock-time as restricted by this specification"?

cconcolato (approved):

index.html
nigelmegitt: Equivalent done in 14589e1.

PR #126, opened 2023-04-06
Draft - Misc editorial fixes

cconcolato:
Close #112

PR #132, opened 2023-04-18
Fixes to pass pubrules (mostly HTML validator) errors

himorin:
Modified:

- removed `section id=""`, and let respec generates id (some hand written IDs were invalid, also the same for ones by respec)
- removed invalid `title` attribute in SVG elements
- fixed nested `<a>` in SVG
- removed deprecated SVG attribute (`contentStyleType`)
- fixed invalid id in SVG (invalid character used)

After this, manual editing on two points are required to pass HTML validator

- issue 66 (closed) has task checkbox, and checkboxes are included with blank id
- TTML2 elements like `<ttp:feature>` are wrongly listed as html element in `Terms defined by reference`

For first one, is it ok to remove this closed issue?
For second one, should be respec bug, but not sure where dereferencing of `&gt;` is actually happens.


index.html
```diff
@@ -2528,10 +2533,11 @@ <h3>#xmlLang-root</h3>
...
     </section> <!-- extensions-->
 
-  <section id="acknowledgements">
+  <section class="appendix">
       <h2>Acknowledgments</h2>
       The editors would like to thank XXX for their contributions to this specification.
   </section>
-
+  <section id="conformance">
+  </section>
```
nigelmegitt, line 2541: This is a breaking change - there's already a conformance section (§6).

index.html
```diff
@@ -364,7 +364,7 @@ <h2>Documentation Conventions</h2>
       </p>
     </section>
 
-  <section id="data-model">
+  <section>
```
nigelmegitt, line 367: The auto-generated id for this is `dapt-data-model-and-corresponding-ttml-syntax` - I much prefer the shorter manual one.

nigelmegitt (changes requested):
I don't mind using auto-generated fragment IDs though it's useful to have the section end comments with the relevant IDs for editing.

But the conformance section change does need fixing please.

index.html
himorin: Ah, thank you for pointing this.
Yes, this was put by accident (and regression by section id="" edits), and let me fix as before.

himorin:
> I don't mind using auto-generated fragment IDs though it's useful to have the section end comments with the relevant IDs for editing.

reverted all comments (removed by regex-like operation, sorry for these)

nigelmegitt (approved):
Thanks, looks good to me.

himorin:
@cconcolato could you kindly review this?

nigelmegitt:
> After this, manual editing on two points are required to pass HTML validator
> 
>     * issue 66 (closed) has task checkbox, and checkboxes are included with blank id

This is removed in #126, should be merged later today.

>     * TTML2 elements like `<ttp:feature>` are wrongly listed as html element in `Terms defined by reference`

I can't see this in the preview (where I don't see them at all), but I do see it on my local machine where the browser shows two empty bullets in that section. When I inspect them in dev tools I see that they are indeed being incorrectly translated from `&lt;ttp:feature&gt;` into `<ttp:feature></ttp:feature` in the HTML. I looked for workarounds but didn't find any so far.

> 
> 
> For first one, is it ok to remove this closed issue? For second one, should be respec bug, but not sure where dereferencing of `&gt;` is actually happens. (filed as [w3c/respec#4430](https://github.com/w3c/respec/issues/4430))

himorin:
> > After this, manual editing on two points are required to pass HTML validator
> > ```
> > * issue 66 (closed) has task checkbox, and checkboxes are included with blank id
> > ```
> 
> This is removed in #126, should be merged later today.

great!

> > ```
> > * TTML2 elements like `<ttp:feature>` are wrongly listed as html element in `Terms defined by reference`
> > ```
> 
> I can't see this in the preview (where I don't see them at all), but I do see it on my local machine where the browser shows two empty bullets in that section. When I inspect them in dev tools I see that they are indeed being incorrectly translated from `&lt;ttp:feature&gt;` into `<ttp:feature></ttp:feature` in the HTML. I looked for workarounds but didn't find any so far.

Yeah,,, so I've already filed an issue as below

> > For second one, should be respec bug, but not sure where dereferencing of `&gt;` is actually happens. (filed as [w3c/respec#4430](https://github.com/w3c/respec/issues/4430))

cconcolato (approved):

nigelmegitt:
The fix for defined terms including HTML character references has been merged to main in respec, at w3c/respec#4431.

PR #135, opened 2023-04-18
Calculation of timing properties for end on both End and Duration are defined

himorin:
I believe we need this fix. Comparing Begin with Begin + Duration and picking the earlier results into Begin always.

Also, I wonder whether this Note: part (and following one also?) should be normative or not. It seems for me, these two notes define exceptional behaviors, which we cannot tell from current normative definitions and also no other normative text says anything about this.

***
<a href="https://pr-preview.s3.amazonaws.com/himorin/dapt/pull/135.html" title="Last updated on Apr 18, 2023, 1:40 PM UTC (cd8694f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/135/a2b8c06...himorin:cd8694f.html" title="Last updated on Apr 18, 2023, 1:40 PM UTC (cd8694f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -568,7 +568,7 @@ table.coldividers td + td { border-left:1px solid gray; }
                 and is relative to the active begin time of the parent object.</li>
               <li>The <dfn>Duration</dfn> property defines the maximum duration of an object.</li>
                 <p class="note">If both an <a>End</a> and a <a>Duration</a> property are present,
-                  the end time is the earlier of <a>Begin</a> and <a>Begin</a> + <a>Duration</a>.</p>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Correct, this is a bug, there's a PR already open at #126 - fix at https://github.com/w3c/dapt/pull/126/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051L571

himorin:
Oh, I could not be noticed that. Thank you for pointing.
closing as dup.

PR #139, opened 2023-04-27
added GHAction for streamlined publication

himorin:
as title

I suppose we'd better to keep three directories in `path`, but if removing some (like `examples`) should be better, please let me know.


himorin:
added W3C_TR_TOKEN, ready to merge and run

himorin:
note. GHaction is working fine.
https://www.w3.org/standards/history/dapt

PR #158, opened 2023-06-14
clarify what spans are possible in a text and how they are handled

### LINKED ISSUE #17: Absence of adaptation information.
Step PS2d generally modifies the word or syllable timing of the text.
In TTAL, this is supported through a separate char offset->time table in each event.
In .srtdub, it is supported by interspersing <begin start=(time offset)/> inside the text.

In TTML, it could be represented by <span begin=(time offset)>text</span>

cconcolato:
close #17

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/158.html" title="Last updated on Feb 15, 2024, 6:40 PM UTC (ca1f7db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/158/8c90163...ca1f7db.html" title="Last updated on Feb 15, 2024, 6:40 PM UTC (ca1f7db)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -927,9 +927,9 @@ table.coldividers td + td { border-left:1px solid gray; }
             <ul>
               <li>The <a>Text</a> of the <a>Script Event</a> is represented by the character content
               of the <code>&lt;p&gt;</code> element and of all of its <code>&lt;span&gt;</code> descendant elements,
-              excluding <code>&lt;metadata&gt;</code> elements and foreign elements,
-              treating <code>&lt;br&gt;</code> elements as new lines,
-              and applying White Space Handling as defined in [[!XML]].
+              after <code>&lt;metadata&gt;</code> elements and foreign elements have been pruned,
+              after replacing <code>&lt;br&gt;</code> elements by line breaks,
+              and after applying White Space Handling as defined in [[!XML]].
                 <aside class="note">The text content of the paragraph can be structured using TTML elements such as
                   <code>&lt;br&gt;</code> or <code>&lt;span&gt;</code>
                   which can include or reference TTML style attributes
--- a/index.html
+++ b/index.html
@@ -925,9 +925,13 @@ table.coldividers td + td { border-left:1px solid gray; }
           </p>
           <p>A <a>Text</a> object is represented in a <a>DAPT Document</a> by a <code>&lt;p&gt;</code> element with the following constraints:</p>
             <ul>
-              <li>The text content of the <code>&lt;p&gt;</code> element is the <a>Text</a> of the <a>Script Event</a>.
-                <aside class="note">The text content of the paragraph can be structured using TTML mark-up such as
-                  <code>&lt;br&gt;</code> or <code>&lt;span&gt;</code> elements
+              <li>The <a>Text</a> of the <a>Script Event</a> is represented by the character content
+              of the <code>&lt;p&gt;</code> element and of all of its <code>&lt;span&gt;</code> descendant elements,
+              excluding <code>&lt;metadata&gt;</code> elements and foreign elements,
+              treating <code>&lt;br&gt;</code> elements as new lines,
+              and applying White Space Handling as defined in [[!XML]].
+                <aside class="note">The text content of the paragraph can be structured using TTML elements such as
+                  <code>&lt;br&gt;</code> or <code>&lt;span&gt;</code>
                   which can include or reference TTML style attributes
                   such as <code>tts:ruby</code> used to alter the layout or styling of
                   sections of text within each paragraph.
@@ -996,11 +1000,14 @@ table.coldividers td + td { border-left:1px solid gray; }
                 This is represented in a <a>DAPT Document</a> by setting the <code>xml:lang</code> and
                 <a><code>daptm:langSrc</code></a>
                 attributes on inner <code>&lt;span&gt;</code> elements.</p>
+                <p class="note"><code>&lt;span&gt;</code> elements can be used to add specific timing
+                as illustrated in [[[#example-10]]] to indicate the timing of the audio rendering
+                of the relevant section of text. Per [[TTML2]], timing of the <code>&lt;span&gt;</code>
+                element is relative to the parent element's computed begin time.</p>
               </li>
               <li>It MAY contain zero or more <code>&lt;audio&gt;</code> elements representing each <a>Audio Recording</a> object.</li>
               <li>It MAY contain zero or more <code>&lt;animate&gt;</code> elements representing each <a>Mixing Instruction</a> object.</li>
             </ul>
-
         </section>
 
         <section>


index.html
```diff
@@ -857,11 +857,17 @@ <h4>Text</h4>
                 <a>Text</a> objects in one language can also contain some words in a different language.
                 This is represented in a <a>DAPT Document</a> by setting the <code>xml:lang</code> attribute on
                 inner <code>&lt;span&gt;</code> elements.</p>
+                <p class="note"><code>&lt;span&gt;</code> elements can be used to add specific styling (e.g. Japanese ruby annotations).</p>
+                <p class="note"><code>&lt;span&gt;</code> elements can be used to add specific timing
+                as illustrated in [[[#example-10]]] to indicate the timing of the audio rendering
+                of the relevant section of text. Per [[TTML2]], timing of the <code>&lt;span&gt;</code>
+                element is relative to the parent element's computed begin time.</p>
+                <p class="note">The text content of a <a>Text</a> object is the concatenated text content
+                of the <code>&lt;p&gt;</code> element and its descendent <code>&lt;span&gt;</code> elements.</p>
```
nigelmegitt, line 866: I think (feel free to check) that it's the _character_ content of the `<p>` element and its descendent `<span>`s - also should we mention white space handling here, since typically consecutive white space is collapsed?

nigelmegitt, line 866: Suggest moving these two up to the end of the first bullet in the list above - here they look like they might be something to do with `xml:lang`.

index.html
```diff
@@ -857,11 +857,17 @@ <h4>Text</h4>
                 <a>Text</a> objects in one language can also contain some words in a different language.
                 This is represented in a <a>DAPT Document</a> by setting the <code>xml:lang</code> attribute on
                 inner <code>&lt;span&gt;</code> elements.</p>
+                <p class="note"><code>&lt;span&gt;</code> elements can be used to add specific styling (e.g. Japanese ruby annotations).</p>
```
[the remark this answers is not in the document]
cconcolato: Sorry, I will remove this one.

index.html
cconcolato: Yes, we should mention white space handling, at least to say that the TTML semantics apply.

cconcolato: In [XML1.0](https://www.w3.org/TR/xml/), "character content" is only mentioned once, with no definition. 
In [TTML2](https://www.w3.org/TR/2018/REC-ttml2-20181108/), it is used 5 times but not linked to any specific definition.
It wasn't clear to me why I proposed it as a note. I now propose to rephrase the first bullet point in this list.

index.html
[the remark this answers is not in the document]
cconcolato: done

index.html
nigelmegitt: I think this is the right term though. For example it is used in XPath Data Model in the overview of [Text Nodes](https://www.w3.org/TR/xpath-datamodel-31/#TextNode):
>Text Nodes encapsulate XML character content.

cconcolato: I changed the text to use "character content"

index.html
```diff
@@ -922,7 +922,9 @@ <h4>Text</h4>
           </p>
           <p>A <a>Text</a> object is represented in a <a>DAPT Document</a> by a <code>&lt;p&gt;</code> element with the following constraints:</p>
             <ul>
-              <li>The <a>Text</a> of the <a>Script Event</a> is represented by the content of the <code>&lt;p&gt;</code> and its descendants, and after applying White Space Handling as defined in [[!XML]].
+              <li>The <a>Text</a> of the <a>Script Event</a> is represented by the character content
+              of the <code>&lt;p&gt;</code> and its descendants, and after applying White Space Handling
```
nigelmegitt, line 926: Not sure if this is over-worrying, but is there a chance someone could parse this as "the character content of the `<p>` plus the (not necessarily character content) descendants of the `<p>`"?

cconcolato: Would "and that of its descendants" work? If not, can you suggest alternative wording?

index.html
```diff
@@ -923,7 +923,7 @@ <h4>Text</h4>
           <p>A <a>Text</a> object is represented in a <a>DAPT Document</a> by a <code>&lt;p&gt;</code> element with the following constraints:</p>
             <ul>
               <li>The <a>Text</a> of the <a>Script Event</a> is represented by the character content
-              of the <code>&lt;p&gt;</code> and its descendants, and after applying White Space Handling
+              of the <code>&lt;p&gt;</code> and of its descendants, and after applying White Space Handling
```
nigelmegitt, line 926: Really sorry, I don't know why I'm thinking of these issues one at a time rather than all together, but there's another issue: not _all_ of its descendants' character content is the Text - for example any descendant of a `<metadata>` child would not count.

I'm not sure quite how to word it, because `<span>`s can have nested `<span>` children themselves, and so on, but it would also be possible (with undefined semantics) to have a `<span>` descendant of a `<metadata>` element or some other foreign namespace element.

There is probably a better way, but here's what I came up with:

```suggestion
              of the <code>&lt;p&gt;</code> element
              and of its <code>&lt;span&gt;</code> element children
              and (recursively) their <code>&lt;span&gt;</code> element children,
              and after applying White Space Handling
```

cconcolato: What about `<br>` elements?

nigelmegitt: Hmm, how about inserting before ", and after applying White Space Handling" an additional line:

>, where `&lt;br&gt;` elements are considered to represent new lines within the character content

?

cconcolato: What about ruby annotation text? How should we consider it? Should we be silent about it?

nigelmegitt: I think we can be silent about it because ruby annotation text is part of the Text, and the mechanism for inserting it uses nested `<span>`s whose content is already included in the definition. Unless your view is that ruby annotations are _not_ part of the Text, @cconcolato ?

cconcolato: Let's stay silent about it. I pushed the change for line breaks.

cconcolato: I try to address your comments (also squashed and rebased against the main branch to resolve conflicts)

nigelmegitt: Is that 0512851 to resolve https://github.com/w3c/dapt/pull/158#discussion_r1455055567 ? I don't think that wording rejects character content children of  `<span>` elements inside foreign namespace elements - I think we agree that such content should not be included in the definition of Text?

cconcolato: I added pruning (but it probably should be rephrased even more, as pruning should happen first). 

I'm not convinced we need all of that text. So far, the spec is written using the following structure: 
"The [part of the DAPT model] is represented by [some syntax in TTML]"
I think it is expressing a one way mapping: how from the model you build the syntax, not how from the syntax you obtain the model. 

Maybe we want that bidirectional mapping? That would require reviewing the rest of the spec (e.g. what if the TTML elements representing a Character, contain more than the few elements specified in the spec).

Assuming, the mapping is one way for now (for consistency with other parts of the spec), the text in this PR should only say that the Text object is represented by a p element, possibly structured in spans, possibly with timing, possibly with interspersed metadata and foreign elements. All the "possibly" parts are covered by notes already.

nigelmegitt (approved):
Thanks, this works for me.

index.html
cconcolato: @nigelmegitt I would be interested in your feedback on the above comment.

nigelmegitt: I'm still thinking about this!

My instinct is that it's fine as is, and maybe we should open a separate issue about this question. I think the key question for implementers is what they should do if they encounter a TTML2 document that claims to be a DAPT document but doesn't exactly match the data model. Are we telling them to implement all relevant TTML2 semantics or only those specifically mentioned in DAPT? Formally I believe it is the set of TTML2 features defined as being required for processors. From that perspective I don't believe we need to say any more about it. 

But happy to discuss further.

cconcolato:
As discussed in the TTWG call https://www.w3.org/2024/02/29-tt-minutes.html#t03,  a new issue #214 was created and this PR can be merged.

PR #159, opened 2023-06-14
add editor's note around metadata rule to gather feedback

### LINKED ISSUE #147: Consider identifying DAPT metadata from other metadata
The spec currently says:
> All `<ttm:agent>` elements SHOULD be contained in the first `<metadata>` element in the `<head>` element.
> NOTE: There can be multiple `<metadata>` elements in the `<head>` element, for example to include [proprietary metadata](https://w3c.github.io/dapt/#proprietary-metadata) but the above recommends that only one is used to define the characters.

The "should" is meant to permit mixing different types of metadata, from different applications. That is fine. But the problem is that is not easy to spot which metadata to ignore and which metadata to process. I wonder if we shouldn't say: if a metadata has an attribute set (e.g. daptm:metadataType='character'), then it is shall contain only the definition of DAPT characters.

cconcolato:
close #147

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/159.html" title="Last updated on Jun 14, 2023, 6:14 PM UTC (e17da71)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/159/3b9e751...e17da71.html" title="Last updated on Jun 14, 2023, 6:14 PM UTC (e17da71)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -688,6 +688,13 @@ table.coldividers td + td { border-left:1px solid gray; }
             <div class="note">There can be multiple <code>&lt;metadata&gt;</code> elements in the <code>&lt;head&gt;</code> element,
             for example to include <a href="#proprietary-metadata">proprietary metadata</a>
             but the above recommends that only one is used to define the characters.</div>
+            <aside class="ednote">The group is considering updating the rule(s) around which metadata element
+            is used to carry DAPT information. The group would like to balance simplicity of implementation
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
LGTM as Editor; as Chair, I class this as editorial so it can be merged without the full 2 week review period.

PR #160, opened 2023-06-14
add note to clarify relationship with SSML

### LINKED ISSUE #121: Clarify how to use SSML with DAPT
Given the overlap between DAPT and SSML, it would be good to have a clarification on how they relate and can be used together (or not). Section https://w3c.github.io/dapt/#foreign-elements-and-attributes could have an example of "proprietary" metadata mixing SSML and DAPT.

cconcolato:
close #121

### DIFF
--- a/index.html
+++ b/index.html
@@ -1207,6 +1207,9 @@ daptm:eventType : string
             whose computed value of <code>tta:rate</code> is <code>none</code>
             has no effect.
             Such an element is not considered to have an associated <a>Synthesized Audio</a>.</p>
+          <p class="note">The <a>Synthesized Audio</a> vocabulary of DAPT overlaps with [[SSML]]. This version
+          of the specification does not specify how [[SSML]] can be either generated from DAPT or embedded
+          into <a>DAPT documents</a>. Future versions of this specification may do so.</p>
         </section>
       </section>
       <section>


index.html
```diff
@@ -1207,6 +1207,9 @@ <h4>Synthesized Audio</h4>
             whose computed value of <code>tta:rate</code> is <code>none</code>
             has no effect.
             Such an element is not considered to have an associated <a>Synthesized Audio</a>.</p>
+          <p class="note">The <a>Synthesized Audio</a> vocabulary of DAPT overlaps with [[SSML]]. This version
+          of the specification does not specify how [[SSML]] can be either generated from DAPT or embedded
+          into <a>DAPT documents</a>. Future versions of this specification may do so.</p>
```
nigelmegitt, line 1212: I like the idea of this note, just minor grammatical/fussy tweaks:

```suggestion
          <p class="note">The semantics of the <a>Synthesized Audio</a> vocabulary of DAPT are derived from equivalent features in [[SSML]]. This version
          of the specification does not specify how other features of [[SSML]] can be either generated from DAPT or embedded
          into <a>DAPT documents</a>. The option to extend [[SSML]] support in future versions of this specification is deliberately left open.</p>
```

Reasons:
* TTML2 states informatively that the semantics of [`tta:pitch`](https://www.w3.org/TR/ttml2/#audio-style-attribute-pitch) and [`tta:speak`](https://www.w3.org/TR/ttml2/#audio-style-attribute-speak) are derived from SSML.
* The _vocabulary_ itself does not overlap with SSML, but the semantics do.
* The gap is with other features of SSML than those two, and it's not necessarily about syntactically matching vocabulary.
* We should avoid normative language (here, "may") from informative text.

nigelmegitt (changes requested):

nigelmegitt (approved):

PR #178, opened 2023-07-31
Try using actions/checkout@v3

nigelmegitt:
Fix build issue caused by previous version of actions/checkout@v2 requiring node 12, which is no longer supported. This version requires node 16, which is supported.

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -15,7 +15,7 @@ jobs:
     name: Build, Validate and Deploy
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: w3c/spec-prod@v2
         with:
           TOOLCHAIN: respec


nigelmegitt:
@himorin this is to fix the run failure at https://github.com/w3c/dapt/actions/runs/5715959932 caused by actions/checkout@v2 using an out-of-date version of node. Appears to work as far as I can tell, and checking the details of the action, the version change is non-breaking - it would allow us to be more specific about which files we check out,  but I don't think we need to do that.

nigelmegitt:
Looks like I misunderstood the cause of the build failure and that #180 has fixed it. Nevertheless, since there is at least a warning about using node 12, and this updates the dependency to rely on node 16 instead, it's probably still worth making this change, unless there's a problem with it.

himorin:
Seems good for me.
Note, I could not find any real sample (as far as I have searched) using ubuntu-22.04 (which is current ubuntu-latest).

himorin (approved):

Issue #148, opened 2023-06-08
Consider identifying the original language on top of the current language


cconcolato:
For dub scripts, in some cases, the document may not contain anymore the original language (for example after recording). It could still be useful to identify the original language (if only to make it clear that the language is not the original). It could be with a `daptm:originalLang` attribute at the tt level.

nigelmegitt:
The original transcript might not all be in one language, so a tt level attribute might not always work. That would suggest that the script event level is more appropriate.

cconcolato:
> The original transcript might not all be in one language, so a tt level attribute might not always work. That would suggest that the script event level is more appropriate.

Just like `xml:lang` in the `originalTranscript`. The meaning of `daptm:originalLang` would simply be: it represents the value that `xml:lang` had in the `originalTranscript`.

nigelmegitt:
If you want to know the original language, why would you not also want to know what the original language text was? If you keep the original transcript text, then you already know the original language.

A secondary consideration: is it important to distinguish between the source language for each translation?
e.g.:

* Original -> Pivot
* Pivot -> Target
* Original -> Target

In this case the last, direct, translation might be higher value than the one that went via the pivot.

nigelmegitt:
Would we allow `daptm:originalLang` to be present on Text `<p>`s additionally, to indicate the language used as the source for translation of that specific Text?

cconcolato:
In case the source had multiple languages?

nigelmegitt:
Oh, yes, that's one reason. The other is to indicate if the translation was direct from the original language or was from a pivot language intermediate translation.

cconcolato:
We could consider langSrc having 3 values: original, translation (meaning target) and pivot; and documents could have up to 3 paragraphs per event.

nigelmegitt:
I'll work on a pull request for this.

nigelmegitt:
I've brought this change into #179 because it is directly related, and reviewing one pull request to resolve this issue and #173 seems to make sense.

Issue #75, opened 2022-11-17
Consider defining restrictions per Script Type


cconcolato:
Some restrictions on the model, e.g. on cardinality #12 or based on profiles #46, could be dependent on the script type, e.g. the presence of Contextual Text. Should we have the generic definition be loose and have additional sections providing restrictions per script type?

nigelmegitt:
In #77 I've made the script-type contents an informative note rather than normative requirements, because I think the flexibility needed probably applies to all script types.

For example see [Translated Dialogue List (PR preview link)](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/77.html#dfn-translated-dialogue-list)

Otherwise we would likely create a requirement for transient non-conformance.

For example, when creating an Original Language Dialogue List, initially there would be no transcribed text, so if we require, say, minimum 1 `div` including minimum 1 `p`, then the first workflow step, of creating effectively an "empty" document, maybe with some pre-defined metadata but no text content, would transiently create a non-conformant document, until at least one Script Event with Text is created. I'd prefer to avoid that.

nigelmegitt:
Close as "won't do"?

cconcolato:
Given that the profile syntax is multi-purpose, I still see value in being clear regarding what's allowed for a particular script type. It could allow workflows to validate incoming documents. Examples of potential validations:
- a dubbing script should not contain audio description content, 
- an dubbing original script should not contain 'translated' text
- dubbing script should not have empty divs, at least one text should be present
- a pre-recording audio description should not contain any `<audio>` elements
- an as-recorded audio description should not contain rate and pitch information
- audio descriptions should not contain characters 
- ...

nigelmegitt:
> * a dubbing script should not contain audio description content,

I'm not sure how you'd identify this?

> * a dubbing original script should not contain 'translated' text

This is already a SHOULD requirement.

> * a dubbing script should not have empty divs, at least one text should be present

This seems overly restrictive to me. It prevents reasonable intermediate production stages, like creating a "shell" with empty timed Script Events where some upstream process has identified that there is dialogue to be dubbed, and that shell acts as a task list.

> * a pre-recording audio description should not contain any `<audio>` elements

Yes, we could add a SHOULD NOT contain `<audio>` constraint to the pre-recording stages.

> * an as-recorded audio description should not contain rate and pitch information

Yes, we could add a SHOULD NOT constraint there.

> * audio descriptions should not contain characters

[...]

Issue #52, opened 2022-09-16
We need think about either collaborative editing or partial editing and whether we need "unfinished" state. Worth an issue.


nigelmegitt:
We need think about either collaborative editing or partial editing and whether we need "unfinished" state. Worth an issue.

_Originally posted by @cconcolato in https://github.com/w3c/dapt/pull/49#discussion_r973503810_

nigelmegitt:
Possibly an issue for a future discussion; not marking as Agenda today because there's not been enough discussion on this issue to warrant it.

nigelmegitt:
Discussed 2023-02-09 with @cconcolato : Agreed to highlight this question somewhere in the specification so that when we go to FPWD we can ask for feedback from the wider world on how important it is to support partial editing.

cconcolato:
Should we add a `daptm:stage = draft | final`?

nigelmegitt:
Maybe `daptm:status = draft | final` but yes, that'd be okay by me. Orthogonal to the script type, i.e. it's reasonable to have a draft or a final version of each script type.

mattsimpson3:
Hi - further to conversations with Nigel, we would suggest a means of tracking changes once something has been edited - so that a delta can readily be produced for reversioning (i.e. a list of points in a script or programme to review and chnage). Frequently the international release of content is tweaked (for music rights, unsuitable language etc) and it would be good to know which sections of the script have been modified. I don't think this needs to be word by word within the DAPT doc, but rather a means of flagging change on sections. Perhaps last modified date / time stamp, as that would allow iteration...?

nigelmegitt:
See also #140.

cconcolato:
Resolving this issue is important to resolve #75. For example, if we don't allow partial documents, there is no reason to have a document with 0 events.

cconcolato:
Current spec says:
https://w3c.github.io/dapt/#script-events
> A [DAPT Script](https://w3c.github.io/dapt/#dfn-dapt-script) MAY contain zero or more [Script Event](https://w3c.github.io/dapt/#dfn-script-event) objects

A final document SHALL have at least one script event, otherwise the document seems useless.

Similarly, https://w3c.github.io/dapt/#dfn-script-event says:
> A Script Event object [...] has the following properties:
Zero or more [Text](https://w3c.github.io/dapt/#dfn-text) objects [...]

A Script Event without Text in a final document does not make sense.

nigelmegitt:
[...]

PR #179, opened 2023-07-31
Rework audio description original and translation languages

### LINKED ISSUE #173: Use of the value "Original" for Text Language Source when it refers to non-dialogue sound
In Section [4.5](https://w3c.github.io/dapt/#text-language-source), it is stated that the value `Original` for the Text Language Source property could be used in case the associated content is a

> representative of non-dialogue sound and is in the Primary Language of the DAPT Script.

If, in a translation, the original text is preserved as shown in [Example 9](https://w3c.github.io/dapt/#example-9), it would then not be possible to include a French description of the non-dialogue sound of the (original) French transcript.

nigelmegitt:
Closes #173 and #148.

Text Language Source updated to remove the link between it and Primary Language. Needed to introduce a new concept of a "nominal" original language for the programme, for the Original language of the Audio Description.

Added notes and examples about how to handle non-nominal original language in-image text in audio descriptions, and the potential need to translate it.

Text Language Source now indicates the source language for the translation.

Tidied some of the existing notes.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/179.html" title="Last updated on Dec 21, 2023, 5:41 PM UTC (aad01da)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/179/f112658...aad01da.html" title="Last updated on Dec 21, 2023, 5:41 PM UTC (aad01da)">Diff</a>

### DIFF
--- a/examples/event-desc-multiple-descType.xml
+++ b/examples/event-desc-multiple-descType.xml
@@ -3,7 +3,7 @@
     <div begin="10s" end="13s">
       <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
       <ttm:desc daptm:descType="plotSignificance">High</ttm:desc>
-      <p>
+      <p xml:lang="en">
         <span>A woman climbs into a small sailing boat.</span>
       </p>
       <p xml:lang="fr" daptm:langSrc="en">
--- a/examples/event-desc-multiple-descType.xml
+++ b/examples/event-desc-multiple-descType.xml
@@ -13,7 +13,7 @@
     <div begin="18s" end="20s">
       <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
       <ttm:desc daptm:descType="plotSignificance">Low</ttm:desc>
-      <p>
+      <p xml:lang="en">
         <span>The woman pulls the tiller and the boat turns.</span>
       </p>
       <p xml:lang="fr" daptm:langSrc="en">
--- a/examples/event-desc-multiple-descType.xml
+++ b/examples/event-desc-multiple-descType.xml
@@ -6,17 +6,17 @@
       <p daptm:langSrc="original">
         <span>A woman climbs into a small sailing boat.</span>
       </p>
-      <p xml:lang="fr" daptm:langSrc="translation">
+      <p xml:lang="fr" daptm:langSrc="en">
         <span>Une femme monte à bord d'un petit bateau à voile.</span>
       </p>
     </div>
     <div begin="18s" end="20s">
       <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
       <ttm:desc daptm:descType="plotSignificance">Low</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
         <span>The woman pulls the tiller and the boat turns.</span>
       </p>
-      <p xml:lang="fr" daptm:langSrc="translation">
+      <p xml:lang="fr" daptm:langSrc="en">
         <span>La femme tire sur la barre et le bateau tourne.</span>
       </p>
     </div>
--- a/examples/event-desc-multiple-descType.xml
+++ b/examples/event-desc-multiple-descType.xml
@@ -3,7 +3,7 @@
     <div begin="10s" end="13s">
       <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
       <ttm:desc daptm:descType="plotSignificance">High</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
         <span>A woman climbs into a small sailing boat.</span>
       </p>
       <p xml:lang="fr" daptm:langSrc="en">
--- a/examples/event-desc.xml
+++ b/examples/event-desc.xml
@@ -2,19 +2,19 @@
   <body>
     <div begin="10s" end="13s">
       <ttm:desc>Scene 1</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
         <span>A woman climbs into a small sailing boat.</span>
       </p>
-      <p xml:lang="fr" daptm:langSrc="translation">
+      <p xml:lang="fr" daptm:langSrc="en">
         <span>Une femme monte à bord d'un petit bateau à voile.</span>
       </p>
     </div>
     <div begin="18s" end="20s">
       <ttm:desc>Scene 1</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
         <span>The woman pulls the tiller and the boat turns.</span>
       </p>
-      <p xml:lang="fr" daptm:langSrc="translation">
+      <p xml:lang="fr" daptm:langSrc="en">
         <span>La femme tire sur la barre et le bateau tourne.</span>
       </p>
     </div>
--- a/examples/event-desc.xml
+++ b/examples/event-desc.xml
@@ -2,7 +2,7 @@
   <body>
     <div begin="10s" end="13s">
       <ttm:desc>Scene 1</ttm:desc>
-      <p>
+      <p xml:lang="en">
         <span>A woman climbs into a small sailing boat.</span>
       </p>
       <p xml:lang="fr" daptm:langSrc="en">
--- a/examples/event-desc.xml
+++ b/examples/event-desc.xml
@@ -11,7 +11,7 @@
     </div>
     <div begin="18s" end="20s">
       <ttm:desc>Scene 1</ttm:desc>
-      <p>
+      <p xml:lang="en">
         <span>The woman pulls the tiller and the boat turns.</span>
       </p>
       <p xml:lang="fr" daptm:langSrc="en">
--- a/figures/sources/class-diagram.puml
+++ b/figures/sources/class-diagram.puml
@@ -44,7 +44,7 @@ together {
 
     Class Text as "**Text**" [[#text]] {
         Text content [[[#dfn-text]]]
-        Text Language Source [[[#text-language-source]]]
+        {field} (optional) Text Language Source [[[#text-language-source]]]
         {field} (optional) Language
         ' {field} (optional) Inline Style Attributes
     }
--- a/figures/sources/class-diagram.puml
+++ b/figures/sources/class-diagram.puml
@@ -16,7 +16,7 @@ Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
     Workflow Type [[[#workflow-type]]]
     Script Type [[[#script-type]]]
     Default Language [[[#default-language]]]
-    Original Languages [[[#original-languages]]]
+    {field} (optional) Original Languages [[[#original-languages]]]
 }
 
 together {
@@ -32,7 +32,6 @@ together {
         {field} {abstract} (optional) End [[[#dfn-end]]]
         {field} {abstract} (optional) Duration [[[#dfn-duration]]]
         {field} (optional) Script Event Type [[[#dfn-script-event-type]]]
-        {field} (optional) Script Event Description [[[#dfn-script-event-description]]]
         {field} (optional) On Screen [[[#on-screen]]]
     }
 
--- a/figures/sources/class-diagram.puml
+++ b/figures/sources/class-diagram.puml
@@ -16,7 +16,7 @@ Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
     Workflow Type [[[#workflow-type]]]
     Script Type [[[#script-type]]]
     Default Language [[[#default-language]]]
-    {field} (optional) Original Languages [[[#original-languages]]]
+    {field} (optional) Text Language Source [[[#text-language-source]]]
 }
 
 together {
--- a/figures/sources/class-diagram.puml
+++ b/figures/sources/class-diagram.puml
@@ -15,7 +15,8 @@ skinparam MinClassWidth 220
 Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
     Workflow Type [[[#workflow-type]]]
     Script Type [[[#script-type]]]
-    Primary Language [[[#primary-language]]]
+    Default Language [[[#default-language]]]
+    Original Language [[[#original-language]]]
 }
 
 together {
--- a/figures/sources/class-diagram.puml
+++ b/figures/sources/class-diagram.puml
@@ -16,7 +16,7 @@ Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
     Workflow Type [[[#workflow-type]]]
     Script Type [[[#script-type]]]
     Default Language [[[#default-language]]]
-    Original Language [[[#original-language]]]
+    Original Languages [[[#original-languages]]]
 }
 
 together {
--- a/index.html
+++ b/index.html
@@ -936,24 +936,36 @@ table.coldividers td + td { border-left:1px solid gray; }
                   <p>In <a>scripts</a> that have very little variation in source language,
                     the <a><code>daptm:langSrc</code></a> attribute can be set on the root element
                     and omitted from <code>&lt;p&gt;</code> elements except where its value
-                    differs.
-                    Care should be taken if using this approach, especially when moving between
+                    differs.</p>
+                  <p>Care should be taken if using this approach, especially when moving between
                     <a>Script Types</a>, because changing it at the root element could affect
                     the interpretation of descendant elements unexpectedly.
-                    Authors can mitigate this risk by explicitly setting <a><code>daptm:langSrc</code></a>
+                    In tools that allow fine-grained control,
+                    authors can mitigate this risk by explicitly setting <a><code>daptm:langSrc</code></a>
                     on all <code>&lt;p&gt;</code> elements.</p>
-                  </p>
-                </aside></li>
+                  <p>Implementers should take care to ensure that,
+                    when changing <a><code>daptm:langSrc</code></a> on an element,
+                    they check down the tree and if appropriate specify the attribute on descendant elements
+                    so that their meaning does not change unintentionally.</p>
+                </aside>
+              </li>
               <li>The <code>&lt;p&gt;</code> element SHOULD have an <code>xml:lang</code> attribute
                 corresponding to the language of the <a>Text</a> object.
-                <p class="note">If a <code>&lt;p&gt;</code> element omits the <code>xml:lang</code> attribute
-                  then its computed language is derived by inheritance from its parent element,
-                  and so forth up to the root <code>&lt;tt&gt;</code> element,
-                  which is required to set the <a>Default Language</a> via its <code>xml:lang</code> attribute.
-                  Care should be taken if changing the <a>Default Language</a> of a <a>DAPT Script</a> in case
-                  doing so affects descendant elements unexpectedly.
-                  Authors can mitigate this risk by explicitly setting <code>xml:lang</code>
-                  on all <code>&lt;p&gt;</code> elements.</p>
+                <aside class="note">
+                  <p>If a <code>&lt;p&gt;</code> element omits the <code>xml:lang</code> attribute
+                    then its computed language is derived by inheritance from its parent element,
+                    and so forth up to the root <code>&lt;tt&gt;</code> element,
+                    which is required to set the <a>Default Language</a> via its <code>xml:lang</code> attribute.</p>
+                  <p>Care should be taken if changing the <a>Default Language</a> of a <a>DAPT Script</a> in case
+                    doing so affects descendant elements unexpectedly.
+                    In tools that allow fine-grained control,
+                    authors can mitigate this risk by explicitly setting <code>xml:lang</code>
+                    on all <code>&lt;p&gt;</code> elements.</p>
+                  <p>Implementers should take care to ensure that,
+                    when changing <code>xml:lang</code> on an element,
+                    they check down the tree and if appropriate specify the attribute on descendant elements
+                    so that their meaning does not change unintentionally.</p>
+                </aside>
                 <pre class="example">
 &lt;div xml:id=&quot;event_3&quot;
      begin=&quot;9663f&quot; end=&quot;9682f&quot; 
--- a/index.html
+++ b/index.html
@@ -697,7 +697,8 @@ languageIdentifier  \\ BCP47 language tag
                   a different language, the <a>Text</a> objects can be translated and marked as <a>Translation</a>.
                   In that case, it is appropriate to mark the transcript as a <a>Translated Transcript</a>.
                   It may also be appropriate, as part of the translation activity, to describe within the translated text
-                  the original language of that text, to inform the audience.
+                  content the original language of that text, to inform the audience,
+                  for example, &quot;A Japanese newspaper headline that means: sailor completes ocean crossing&quot;.
                 </aside>
                 </li>
               <li>
@@ -971,8 +972,8 @@ languageIdentifier  \\ BCP47 language tag
             <ol>
               <li>the same language as the dialogue that it represents in the original programme audio;</li>
               <li>a transcription of text visible in the programme video, in the same language as that text;</li>
-              <li>representative of non-dialogue sound;</li>
-              <li>a description of the scene in the programme video.</li>
+              <li>an untranslated representative of non-dialogue sound;</li>
+              <li>an untranslated description of the scene in the programme video.</li>
             </ol>
           </p>
           <aside class="note">The language of an <a>Original</a> <a>Text</a> object
@@ -1013,7 +1014,6 @@ languageIdentifier  \\ BCP47 language tag
                     then its computed value is derived by inheritance from its parent element,
                     and so forth up to the root <code>&lt;tt&gt;</code> element.</p>
                   <p>In <a>scripts</a> that have very little variation in source language,
-                    for example audio description scripts,
                     the <a><code>daptm:langSrc</code></a> attribute can be set on the root element
                     and omitted from <code>&lt;p&gt;</code> elements except where its value
                     differs.
--- a/index.html
+++ b/index.html
@@ -374,7 +374,9 @@ table.coldividers td + td { border-left:1px solid gray; }
           Note that in this case, to reflect that most of the audio description content
           transcribes the video image where there is no inherent language,
           the <a>Text Language Source</a>, represented by the <a><code>daptm:langSrc</code></a> attribute,
-          is set to the empty string at the top level of the document:</p>
+          is set to the empty string at the top level of the document.
+          It would be semantically equivalent to omit the attribute altogether,
+          since the default value is the empty string:</p>
         <pre class="example"
           data-include="examples/intro-times-and-text.xml"
           data-include-format="text">
@@ -421,8 +423,10 @@ table.coldividers td + td { border-left:1px solid gray; }
         </pre>
 
         <h4>Dubbing Examples</h4>
-        <p>From the basic structure of <a href="#example-1"></a>, a transcription of the audio programme produces an original language dubbing <a>script</a>,
-          which can look as follows. No specific style or layout is defined, and here the focus is on the transcription of the dialogs.
+        <p>From the basic structure of <a href="#example-1"></a>, 
+          transcribing the audio produces an original language dubbing <a>transcript</a>,
+          which can look as follows.
+          No specific style or layout is defined, and here the focus is on the transcription of the dialogue.
           Characters are identified within the <code>&lt;metadata&gt;</code>.
           Note that the language and the <a>text language source</a> are defined using
           <code>xml:lang</code> and <a><code>daptm:langSrc</code></a> respectively,
@@ -437,7 +441,10 @@ table.coldividers td + td { border-left:1px solid gray; }
         in this case the original text is preserved. The main document's <a>default language</a> is changed to indicate
         that the focus is on the translated language.
         The combination of the <code>xml:lang</code> and <a><code>daptm:langSrc</code></a> attributes are used
-        to mark the text as being original or translated:</p>
+        to mark the text as being original or translated.
+        In this case, they are present on both the <code>&lt;tt&gt;</code> and <code>&lt;p&gt;</code>
+        elements to make the example easier to read, but it would also be possible to omit
+        them in some cases, making use of the inheritance model:</p>
         <pre class="example"
           data-include="examples/intro-original-language-with-dub-language.xml"
           data-include-format="text">
@@ -1024,9 +1031,6 @@ daptm:langSrc
             <li>The default value is the empty string.</li>
             <li>It <em>applies</em> to <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code> elements.</li>
             <li>It MAY be specified on the following elements: <code>&lt;tt&gt;</code>, <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code>.</li>
-            <li>If it is present on an element,
-              the computed value is the specified value.
-            </li>
             <li>The inheritance model of <a><code>daptm:langSrc</code></a> is as follows:
               <ul>
                 <li>If it is present on an element,
--- a/index.html
+++ b/index.html
@@ -1046,25 +1046,47 @@ languageIdentifier  \\ BCP47 language tag
         <section>
           <h4>Text Language Source</h4>
           <p>The <dfn>Text Language Source</dfn> property is an annotation indicating the source language of
-            a <a>Text</a> object that represents content with an inherent language.
-            If it differs from the language of the <a>Text</a> object,
+            a <a>Text</a> object, if applicable, or that the source content had no inherent language.</p>
+          <p>If it is not empty and differs from the language of the <a>Text</a> object,
             that implies that it is a representation in another language,
             referred to as &quot;translation&quot;;
             conversely if it is the same that implies that the language of the <a>Text</a> is <a>Original</a>.</p>
-          <p><a>Text Language Source</a> properties SHOULD NOT be present on <a>Text</a> objects
-            that represent content without an inherent language,
+          <p>An empty <a>Text Language Source</a> property on a <a>Text</a> object
+            indicates that the <a>Text</a> represents content without an inherent language,
             such as untranslated descriptions of a visual scene or
             captions representing non-dialogue sounds.</p>
           <p>The <a>Text Language Source</a> property is represented in a <a>DAPT Document</a> by a <dfn><code>daptm:langSrc</code></dfn> attribute
-            whose value MUST be a language identifier as defined by [[BCP47]].</p>
-          <p>If the value of <a><code>daptm:langSrc</code></a> differs from the computed <code>xml:lang</code> then it indicates
-            that the <a>Text</a> is a <a>translation</a>, and is the language from which the <a>Text</a> was translated.
-            Otherwise, if they are the same, then the <a>Text</a> is <a>Original</a>
-            and sourced from content with an inherent language.</p>
-          <p>If a <a>Text</a> object does not have a <a><code>daptm:langSrc</code></a> attribute then it is <a>Original</a>
+            whose value MUST be an empty string or a language identifier as defined by [[BCP47]].</p>
+          <p>The default value of <a><code>daptm:langSrc</code></a> is the empty string.</p>
+          <div class="exampleInner">
+            <pre class="language-abnf">
+daptm:langSrc
+: &lt;empty-string&gt; | &lt;language-identifier&gt;
+
+&lt;empty-string&gt;
+: ""                    # default
+
+&lt;language-identifier&gt;   # valid BCP-47 language tag
+            </pre>
+          </div>
+          <p>The <a><code>daptm:langSrc</code></a> <em>applies</em> to <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code> elements.</p>
+          <p>If no <a><code>daptm:langSrc</code></a> attribute is present on an element,
+            the computed value of the attribute on that element is
+            the computed value of the same attribute on the element's parent, if present,
+            or the default value otherwise.</p>
+          <p class="note">The inheritance model of <a><code>daptm:langSrc</code></a> is intended to match
+            the inheritance model of the <code>xml:lang</code> attribute [[XML]].</p>
+          <p>If the computed value of <a><code>daptm:langSrc</code></a> is the empty string then it indicates
+            that the <a>Text</a> is <a>Original</a>
             and sourced from content without an inherent language.</p>
-          <p>If a <code>&lt;span&gt;</code> element does not have a <a><code>daptm:langSrc</code></a> attribute then
-            its <a>Text Language Source</a> is as defined by its parent element.</p>
+          <p>Otherwise, if the computed value of <a><code>daptm:langSrc</code></a> is not empty and
+            differs from the computed value of <code>xml:lang</code>,
+            then it indicates that the <a>Text</a> is a <a>translation</a>,
+            and is the language from which the <a>Text</a> was translated.</p>
+          <p>Otherwise, if the computed value of <a><code>daptm:langSrc</code></a> is not empty and
+            is the same as the computed value of <code>xml:lang</code>,
+            then it indicates that the <a>Text</a> is <a>Original</a>
+            and sourced from content with an inherent language.</p>
           <p class="note">An example of the usage of <a>Text Language Source</a> in a document is present in the <a>Text</a> section.</p>
           <aside class="example">
             <table class="simple">
@@ -1098,13 +1120,13 @@ languageIdentifier  \\ BCP47 language tag
                   <td>Video image (non text)</td>
                   <td><em>none</em></td>
                   <td><code>en</code></td>
-                  <td><em>absent</em></td>
+                  <td><em>empty</em></td>
                 </tr>
                 <tr>
                   <td>Sound effect</td>
                   <td><em>none</em></td>
                   <td><code>en</code></td>
-                  <td><em>absent</em></td>
+                  <td><em>empty</em></td>
                 </tr>
                 <tr>
                   <td>Dialogue in one of the <a>original languages</a></td>
--- a/index.html
+++ b/index.html
@@ -838,8 +838,7 @@ table.coldividers td + td { border-left:1px solid gray; }
             <li>Zero or more <a>Character Identifiers</a> indicating the <a>Characters</a> involved in this <a>Script Event</a>.
               <p class="note">While typically, a <a>Script Event</a> corresponds to one single <a>Character</a>, there are cases where multiple <a>characters</a> can be associated with a <a>Script Event</a>. This is when all <a>Characters</a> speak the same text at the same time.</p>
             </li>
-            <li><p>Zero or more <a>Text</a> objects, each representing either the <a>Original</a> <a>transcript</a> or
-              a <a>Translation</a>.</p>
+            <li><p>Zero or more <a>Text</a> objects, each being either <a>Original</a> or <a>Translation</a>.</p>
               <p class="note">Empty <a>Text</a> objects can be used to indicate explicitly that there is no text content.
                 It is recommended that empty <a>Text</a> objects are not used as a workflow placeholder to indicate incomplete work.</p></li>
             <li>zero or more <a>Script Event Description</a> objects, each being a human-readable description of the <a>Script Event</a>.</li>
--- a/index.html
+++ b/index.html
@@ -553,13 +553,13 @@ table.coldividers td + td { border-left:1px solid gray; }
                 <p>When the <code>daptm:scriptType</code> value is <code>originalTranscript</code>,
                 the document is a literal transcription of the dialogue and on-screen text in their inherent spoken/written language(s),
                 or of non-dialogue sounds and visual content in the <a>Original Language</a>.</p>
-                <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Original</a>
+                <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain
+                  <a>Original</a> language <a>Text</a> objects
                   and SHOULD NOT contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Translation</a>.</p>
+                  <a>Text Language Source</a> indicates that they are <a>Translations</a>.</p>
               <aside class="example">If a programme contains dialogue in English and Hebrew,
                 the <a>Original Language Transcript</a> will contain some <a>Script Events</a> in English and some in Hebrew,
-                all marked as <a>Original</a> <a>Text Language Source</a>.
+                all containing <a>Original</a> language <a>Text</a>.
                 None of the <a>Script Events</a> is marked as <a>Translation</a>.</aside>
               </li>
               <li>
@@ -569,15 +569,18 @@ table.coldividers td + td { border-left:1px solid gray; }
                 <p>It can be adapted to produce a <a>Pre-Recording Script</a>,
                   and/or used as the basis for a further translation into the <a>Target Recording Language</a>.</p> 
                 <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Translation</a>.
-                  They MAY also contain Text whose language is <a>Original</a>.</p>
+                  <a>Text Language Source</a> is set to the source language from which they were translated,
+                  which also indicates that they are <a>Translations</a>.
+                  They MAY also contain <a>Original</a> <a>Text</a> objects.</p>
                 <aside class="note">A <a>Script Event</a> in this type of <a>transcript</a> has a 
                   <a>Translation</a> <a>Text</a> object in the <a>Target Recording Language</a>
                   and can also contain an <a>Original</a> <a>Text</a> object, for context,
                   to assist in the adaption process.
                   It can also have <a>Translation</a> <a>Text</a> objects in other languages,
                   for example used as intermediate (&quot;pivot&quot;) translations,
-                  or as additional context to assist in the adaptation process.</aside>
+                  or as additional context to assist in the adaptation process.
+                  Each <a>Translation</a> <a>Text</a> object indicates its source language
+                  using the <a>Text Language Source</a> property.</aside>
                 <aside class="example">If a programme contains dialogue in English and Hebrew,
                   the French <a>Translated Transcript</a> will contain at least the translation in French of all <a>Script Events</a>.
                   It may still retain text content in Hebrew and English to assist further processing.</aside>
@@ -587,7 +590,7 @@ table.coldividers td + td { border-left:1px solid gray; }
                   <a>Original Language</a>, the <a>Text</a> objects can be translated and marked as <a>Translation</a>.
                   In that case, it is appropriate to mark the transcript as a <a>Translated Transcript</a>.
                   It may also be appropriate, as part of the translation activity, to describe within the translated text
-                  the original inherent language of the text, to inform the audience.
+                  the original language of that text, to inform the audience.
                 </aside>
                 </li>
               <li>
@@ -605,15 +608,20 @@ table.coldividers td + td { border-left:1px solid gray; }
                 <aside class="note">The <a>Script Type</a> of a <a>DAPT Script</a> cannot necessarily be detected by inspecting the text content of the document.
                 For example, the adaptation of a <a>Translated Transcript</a> into a <a>Pre-recording Script</a>
                 can consist in replacing some words in the text content of a <a>Script Event</a> without altering the rest of the document.
-                In either case, <a>Text</a> objects that are translations have <a>Text Language Source</a> properties set to <a>Translation</a>.</aside>
+                In either case, <a>Text</a> objects that are <a>Translations</a>
+                have <a>Text Language Source</a> properties
+                set to the source language from which they were translated.</aside>
                 
-                <aside class="note">The <a>Text</a> objects in Audio Description <a>Script Events</a> are each expected to have
-                  a <a>Text Language Source</a> property of <a>Original</a> if they represent non-textual visual elements of the scene,
+                <aside class="note">The <a>Text</a> objects in Audio Description <a>Script Events</a> are not expected to have
+                  a <a>Text Language Source</a> property if they represent non-textual visual elements of the scene,
+                  i.e. they are <a>Original</a>
                   and their language is the <a>Original Language</a>, 
                   or, when representing in-image text,
                   if they are in the same language as that text.
                   If audio description <a>scripts</a> are translated,
-                  their translations would be marked as <a>Translation</a>.</aside>
+                  their translations would be marked as <a>Translation</a>
+                  and those translation <a>Text</a> objects would indicate
+                  the source language using the <a>Text Language Source</a> property.</aside>
               </li>
               <li>
                 <dfn>As-recorded Script</dfn>:
@@ -626,7 +634,8 @@ table.coldividers td + td { border-left:1px solid gray; }
                   They MAY also contain links to audio and mixing instructions for the purpose of producing an audio track incorporating the recordings.</p>
                 <aside class="note">Even though the output of the dubbing workflow is in the same language as those
                 <a>Text</a> objects with the matching language,
-                the <a>Text Language Source</a> of those Text objects remains <a>Translation</a>
+                the <a>Text Language Source</a> is retained,
+                indicating the source language of those <a>Translation</a> <a>Text</a> objects 
                 since they differ from the untranslated language of the dialogue.</aside>
               </li>
             </ul>
@@ -658,7 +667,9 @@ table.coldividers td + td { border-left:1px solid gray; }
             When multiple languages are used, the <a>Default Language</a> can correspond to the language of the majority of <a>Script Events</a>,
             to the language being spoken for the longest duration, or to the language arbitrarily chosen by the author.</p>
           <p class="note">The language of text content is marked either as being either <a>Original</a>  or
-            <a>Translation</a> independently of the <a>Default Language</a>, using the <a>Text Language Source</a> property.</p>
+            <a>Translation</a> independently of the <a>Default Language</a>,
+            using the absence or presence of the <a>Text Language Source</a> property,
+            and its value.</p>
           <aside class="example">An <a>Original Language Transcript</a> of dialogue is prepared for a video
             containing dialogue in Danish and Swedish.
             The <a>Default Language</a> is set to Danish by setting <code>xml:lang="da"</code> on the <code>&lt;tt&gt;</code> element.
@@ -667,7 +678,7 @@ table.coldividers td + td { border-left:1px solid gray; }
             <a>Script Events</a> that contain Danish <a>Text</a> can set the <code>xml:lang</code> attribute
             or omit it, since the inherited language is the <a>Default Language</a> of the document.
             In both cases the Script Events' <a>Text</a> objects are <code>&lt;p&gt;</code> elements that represent untranslated text
-            and therefore set their <a>Text Language Source</a> to <a>Original</a> by setting <code>daptm:langSrc="original"</code>.</aside>
+            and therefore omit the <code>daptm:langSrc</code> attribute, implying that they are in the <a>Original</a> language.</aside>
         </section>
 
         <section>
@@ -703,14 +714,14 @@ table.coldividers td + td { border-left:1px solid gray; }
           </p>
           <aside class="example">
             An <a>Original Language Transcript</a> of the video image is prepared for a video
-            containing dialogue in Danish and Swedish, and has it's <a>Original Language</a> set to Danish.
+            containing dialogue in Danish and Swedish, and has its <a>Original Language</a> set to Danish.
             The video features text written in Danish and Swedish.
             The <a>Default Language</a> is set to Danish.
             <a>Script Events</a> that describe non-textual parts of the scene are written in Danish
-            and have their <a>Text Language Source</a> set to Original by setting <code>daptm:langSrc="original"</code>.
+            and have their <a>Text Language Source</a> set to <a>Original</a> by omitting the <code>daptm:langSrc</code> attribute.
             <a>Script Events</a> that describe Swedish text in the video image are written in Swedish
-            and have their <a>Text Language Source</a> set to Original and their language set to Swedish
-            by setting <code>xml:lang="sv"</code> on the <code>&lt;p&gt;</code> element.
+            and have their <a>Text Language Source</a> set to <a>Original</a> and their language set to Swedish
+            by omitting the <code>daptm:langSrc</code> attribute and setting <code>xml:lang="sv"</code> on the <code>&lt;p&gt;</code> element.
           </aside>
         </section>
 
@@ -870,7 +881,7 @@ table.coldividers td + td { border-left:1px solid gray; }
             <li>Zero or more <a>Character Identifiers</a> indicating the <a>Characters</a> involved in this <a>Script Event</a>.
               <p class="note">While typically, a <a>Script Event</a> corresponds to one single <a>Character</a>, there are cases where multiple <a>characters</a> can be associated with a <a>Script Event</a>. This is when all <a>Characters</a> speak the same text at the same time.</p>
             </li>
-            <li><p>Zero or more <a>Text</a> objects, each representing either the untranslated (<a>Original</a>) <a>transcript</a> or
+            <li><p>Zero or more <a>Text</a> objects, each representing either the <a>Original</a> (untranslated) <a>transcript</a> or
               <a>Translations</a> in other languages.</p>
               <p class="note">Empty <a>Text</a> objects can be used to indicate explicitly that there is no text content.
                 It is recommended that empty <a>Text</a> objects are not used as a workflow placeholder to indicate incomplete work.</p></li>
@@ -944,7 +955,14 @@ table.coldividers td + td { border-left:1px solid gray; }
                   <code>tts:direction</code> attribute on <code>&lt;span&gt;</code> elements.
                   Similarly metadata can be added using attributes or <code>&lt;metadata&gt;</code> elements.</aside>
               </li>
-              <li>The <code>&lt;p&gt;</code> element MUST have one <a>Text Language Source</a> property.</li>
+              <li>If the <a>Text</a> is a <a>Translation</a>,
+                the <code>&lt;p&gt;</code> element MUST have one
+                <code>daptm:langSrc</code> attribute representing the <a>Text Language Source</a> property,
+                indicating the source language used as the basis of the translation.</li>
+              <li>If the <a>Text</a> is <a>Original</a>,
+                the <code>&lt;p&gt;</code> element MUST either not have a <code>daptm:langSrc</code> attribute or
+                have a <code>daptm:langSrc</code> attribute whose value is the computed value of <code>xml:lang</code>
+                for the <code>&lt;p&gt;</code> element.</li>
               <li>The <code>&lt;p&gt;</code> element SHOULD have an <code>xml:lang</code> attribute corresponding to the language of the <a>Text</a> object.
                 <p class="note">If a <code>&lt;p&gt;</code> element omits the <code>xml:lang</code> attribute then its computed language
                 is derived by inheritance from its parent element, and so forth up to the root <code>&lt;tt&gt;</code> element,
@@ -956,8 +974,8 @@ table.coldividers td + td { border-left:1px solid gray; }
 &lt;div xml:id=&quot;event_3&quot;
      begin=&quot;9663f&quot; end=&quot;9682f&quot; 
      ttm:agent=&quot;character_3&quot;&gt;
-  &lt;p xml:lang=&quot;pt-BR&quot; daptm:langSrc=&quot;original&quot; &gt;Voc&#xea; vai ter.&lt;/p&gt;
-  &lt;p xml:lang=&quot;fr&quot; daptm:langSrc=&quot;translation&quot; &gt;Bah, il arrive.&lt;/p&gt;
+  &lt;p xml:lang=&quot;pt-BR&quot;&gt;Voc&#xea; vai ter.&lt;/p&gt;
+  &lt;p xml:lang=&quot;fr&quot; daptm:langSrc=&quot;pt-BR&quot;&gt;Bah, il arrive.&lt;/p&gt;
 &lt;/div&gt;
                 </pre>
                 <p class="note">In some cases, a single section of untranslated dialogue can contain text in more than one language.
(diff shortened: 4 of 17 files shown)


andreastai (commented):
The introduction of the term `nominal original language` could make it difficult to understand when to use `original` or `translation`. Currently, there is no definition of what distinguishes a `nominal original language` from an `original language`.

nigelmegitt:
I've been thinking about how to resolve the difficulties you mention @andreastai and it's not straightforward. My thinking so far, based on this PR as it is now:

* "Primary Language" is a concept given too much importance - really it's just a default language for the document, and doesn't carry any other significance.
* When non-dialogue sounds or non-text video image contents are described, we need a concept and a term for expressing what language they are described in, in the "original" transcript.

I'm tempted to downgrade "Primary Language" to "default language" as a term (but not really make any other changes to it), and to introduce something like "Non-verbal Content Original Language" which is a bit of a wordy term, but doesn't need to be used too often, hopefully.

I'm envisaging a specific attribute on the `tt` level for establishing the "Non-verbal Content Original Language", which would not change even if the descriptive content is translated. Then this slightly worrying sounding "nominal original language" would be an explicitly defined thing we could reference.

Does that sound like a good way forward?

nigelmegitt:
I've made a change to this PR roughly matching the intent of https://github.com/w3c/dapt/pull/179#issuecomment-1662573116 that no longer requires or references some invented "nominal" original audio language:

* Rename Primary Language to Default Language
* Introduce Original Language as the mandatory `daptm:originalLang`
* Reference this Original Language within the definition text for Original text language source

I will check the extension feature set matches these requirements too, will likely need another commit and push.

andreastai:
Thanks for your attempt to resolve the issues @nigelmegitt. After a quick first review, I think this formally could work. But we still have the issue that it makes the use of the `Text Language Source` very difficult to understand. Following your proposed concept I think the introduced property "OriginalLanguage" would be better named in a way as proposed in https://github.com/w3c/dapt/pull/179#issuecomment-1662573116 such as `Non-verbal Content Original Language`. 

But the more I think about it I wonder if for non-verbal content the property Text Language Source should be used at all. It should be considered to not identity the language source for non-verbal content or to add another attribute to Text Objects such as `Non-Verbal Content Language Source`.

nigelmegitt:
See discussion logged at https://github.com/w3c/dapt/issues/173#issuecomment-1664221491 for suggestions for how to proceed with this.

nigelmegitt:
Also see https://github.com/w3c/dapt/issues/148#issuecomment-1715864388 for an additional relevant change relating to this.

cconcolato (commented):
Text language source seems to be optional but it does not appear so in the diagram.

andreastai (commented):
Thanks, @nigelmegitt The use of the Text Language Source property is clear. However, the name and usage of the Original Source property of the DAPT Script is confusing. The name does not reflect that it only applies to content that does not have an inherent language (e.g. non-dialogue sounds). Furthermore, I wonder if it would be sufficient to just use the Text Language Source property for content without an inherent language to indicate if the text content was translated and if so from which language.

nigelmegitt:
Thanks @andreastai I can see why "Original Language" is a confusing name for text describing content that does not have an inherent language. Is your proposal to treat Text whose Text Language Source is the same as its Language as a special case to represent that the text was authored to describe that content?

For example:

```xml
<div xml:id="event_3"
     begin="9663f" end="9682f">
  <p xml:lang="en" daptm:langSrc="en">The computer screen turns blue.</p>
</div>
```

andreastai:
> Is your proposal to treat Text whose Text Language Source is the same as its Language as a special case to represent that the text was authored to describe that content?
> 
> For example:
> 
> ```
> <div xml:id="event_3"
>      begin="9663f" end="9682f">
>   <p xml:lang="en" daptm:langSrc="en">The computer screen turns blue.</p>
> </div>
> ```

I would propose not using specific tagging for descriptions of content without an inherent language if it was described just using that content and not a previously existing text, e.g.

```xml
 <div xml:id="event_3"
      begin="9663f" end="9682f">
   <p xml:lang="en">The computer screen turns blue.</p>
 </div
```

But if this is now translated into German based on the English description the `daptm:langSrc` attribute could be used:


```xml
 <div xml:id="event_3"
      begin="9663f" end="9682f">
   <p xml:lang="de" daptm:langSrc="en">Der Computerbildschirm wird blau.</p>
 </div
```

examples/event-desc-multiple-descType.xml
```diff
@@ -6,17 +6,17 @@
       <p daptm:langSrc="original">
```
cconcolato, line 6: shouldn't the langSrc attribute also be removed here?

nigelmegitt: Yes, fixing

nigelmegitt: Done in 932006b

nigelmegitt:
> I would propose not using specific tagging for descriptions of content without an inherent language if it was described just using that content and not a previously existing text,

@andreastai Thanks, I think I understand. Source content with an inherent language has that language marked up in `daptm:langSrc`, and source content without an inherent language omits the attribute. That's close to what we have in this pull request but will need some changes. I think it may only need minor tweaks. Enumerating the possibilities with an example where original language is English:

| Transcript source | Inherent language of what's being transcribed | `xml:lang` | `daptm:langSrc` |
|--------|--------|--------|--------|
| Text in video image in original language | English | `en` | `en` |
| Text in video image in non-original language | German | `de` | `de` |
| Video image (non text) | _none_ | `en` | _absent_ |
| Sound effect | _none_ | `en` | _absent_ |
| Dialogue in original language | English | `en` | `en` |
| Dialogue in non-original language | French | `fr` | `fr` |

Then whenever these pieces of text are translated, the translations have their `daptm:langSrc` set to the computed `xml:lang` of the source for the translation, and at that point the values of `xml:lang` and `daptm:langSrc` differ.

Also, in this proposal there is no longer any particular need to record the document's Original Language, because the original language of the transcript content can always be derived.

Does that match your proposal?

nigelmegitt:
Changes as discussed in https://github.com/w3c/dapt/pull/179#issuecomment-1804198716 now made in 84f9e0c. Hopefully this makes the handling of content in multiple languages, and the handling of transcripts of content without an inherent language much clearer, and removes the complex dependency on Original Language, which is now optional.

@andreastai @cconcolato please re-review when you can.

nigelmegitt:
I would like to merge this on Nov 30 if there are positive reviews and no unresolved objections.

index.html
```diff
@@ -500,19 +502,31 @@ <h4>Script Type</h4>
                 <p>It can be adapted to produce a <a>Pre-Recording Script</a>,
                   and/or used as the basis for a further translation into the <a>Target Recording Language</a>.</p> 
                 <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Translation</a>.
-                  They MAY also contain <a>Original</a> language <a>Text</a>.</p>
+                  <a>Text Language Source</a> is set to the source language from which they were translated,
+                  and whose language is set to their actual language,
+                  indicating that they are <a>Translations</a>.
+                  They MAY also contain <a>Original</a> <a>Text</a> objects.</p>
```
cconcolato, line 508: Should we say that:
* All events SHALL have at least one text object with the same language. 
* That language MAY or MAY NOT be an original language. 
* Other text objects MAY exist marked with an (other) original language.

For example if the original languages are English and French, if the translation language is French, all events will include French text objects and may include English text objects. But if the translated language is German, all events will include German text objects and may include English and French text objects.

index.html
```diff
@@ -522,20 +536,25 @@ <h4>Script Type</h4>
                 <p><a>Script Events</a> in this type of <a>script</a> SHOULD contain <a>Text</a> objects
```
cconcolato, line 536: Also wondering why we have a `SHOULD` and not a `SHALL`. I would suggest closing https://github.com/w3c/dapt/issues/52 by saying that representing unfinished documents is out of scope, and using the SHALL here.

index.html
```diff
@@ -561,35 +581,113 @@ <h4>Script Type</h4>
...
+            used in the workflow.
           </p>
 
-            <p>The <a>Primary Language</a> is represented in a <a>DAPT Document</a> by the following structure and constraints:
+          <p>The <a>Default Language</a> is represented in a <a>DAPT Document</a> by the following structure and constraints:
             <ul>
               <li>the <code>xml:lang</code> attribute MUST be present on the <code>&lt;tt&gt;</code> element and its value MUST NOT be empty.</li>
             </ul>
           </p>
           <p class="note">All text content in a <a>DAPT Script</a> has a specified language.
```
cconcolato, line 599: Should it be "All text content ... have a specified language"?

cconcolato (commented):
Thanks for this extensive Pull Request! 

When I read it linearly, even if I had the initial context, it was hard to understand some parts as some concepts were defined afterwards. For example, 4.1.2 makes heavy use of concepts in 4.1.3, 4.1.4 and even 4.5. I'm not sure how to fix that. I would recommend creating a section around "Language considerations/overview" that include all or some of 4.5, 4.1.3. and 4.1.4 and put that section early in the document, but it might not play nice with the current document structure that matches model objects or properties.

Also, I wonder what we gain by permitting languages that are briefly spoken not to be in the original language list? If all original languages were mandated in the list, you could easily tell all the languages in the document by looking at the document header, and avoid the edge case of a langSrc not listed in the originalLangs list.

figures/sources/class-diagram.puml
```diff
@@ -15,7 +15,8 @@ skinparam MinClassWidth 220
 Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
     Workflow Type [[[#workflow-type]]]
     Script Type [[[#script-type]]]
-    Primary Language [[[#primary-language]]]
+    Default Language [[[#default-language]]]
+    Original Language [[[#original-language]]]
```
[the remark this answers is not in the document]
nigelmegitt: I don't mind renaming it Original Languages.

index.html
```diff
@@ -209,7 +207,7 @@ <h4>Dubbing <a>scripts</a></h4>
           <p>The <dfn>dubbing</dfn> process which consists in creating a <a>dubbing script</a>
             is a complex, multi-step process involving:</p>
           <ul>
-            <li>Transcribing and timing the dialogue in the original language from a completed programme to create a <a>transcript</a>;</li>
+            <li>Transcribing and timing the dialogue in its own language from a completed programme to create a <a>transcript</a>;</li>
```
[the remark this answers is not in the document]
nigelmegitt: I changed it because "original language" is a defined term, and it would be confusing to use it here in a non-defined term sense.

index.html
```diff
@@ -292,7 +290,9 @@ <h4>Basic document structure</h4>
           and the <a><code>ttp:contentProfiles</code></a> attribute indicates that it adheres to the DAPT <a>content profile</a> defined in this specification.</li>
           <li>The <code>daptm:workflowType</code> attribute indicates the type of workflow.</li>
           <li>The <code>daptm:scriptType</code> attribute indicates the type of <a>transcript</a> or <a>script</a>
-          but in this empty example, it is not relevant, since only the structure of the document is shown.</li>
+          but in this empty example, it is not relevant, as only the structure of the document is shown.</li>
+          <li>The <code>daptm:originalLangs</code> attribute indicates the languages used
```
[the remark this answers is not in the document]
nigelmegitt: Sure, I can put the brackets in. I think I'll also switch "as only" back to "since only" because it's clearer.

index.html
```diff
@@ -483,14 +484,15 @@ <h4>Script Type</h4>
               <li>
                 <dfn>Original Language Transcript</dfn>:
                 <p>When the <code>daptm:scriptType</code> value is <code>originalTranscript</code>,
-                the document is a literal transcription of the dialogue and on-screen text in their original spoken/written language(s).</p>
-                <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Original</a>
+                the document is a literal transcription of the dialogue and/or on-screen text in their inherent spoken/written language(s),
```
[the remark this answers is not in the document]
nigelmegitt: "original" is now a defined term - I think I was being careful about how I used it.

index.html
```diff
@@ -483,14 +484,15 @@ <h4>Script Type</h4>
...
+                or of non-dialogue sounds and non-linguistic visual content.</p>
+                <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain
+                  <a>Original</a> language <a>Text</a> objects
                   and SHOULD NOT contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Translation</a>.</p>
+                  <a>Text Language Source</a> indicates that they are <a>Translations</a>.</p>
```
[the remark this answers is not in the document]
nigelmegitt: That's not a change in this PR. If I recall correctly it is because the Script Events may contain no Text objects at all, in the initial stages of authoring AD for example.

index.html
nigelmegitt: I don't follow: we don't have explicit metadata indicating what is the Target Recording Language, so I don't think that those rules would be testable. Isn't the text in this PR saying what you mean though? Possibly there's a requirement I haven't captured or understood so far.

index.html
nigelmegitt: My instinct is that I'd rather not do that, and that it'd be somewhat unfriendly, but I'd prefer to get implementation and usage feedback before deciding to do that.

index.html
nigelmegitt: I think "content" is singular not plural.

nigelmegitt:
> Thanks for this extensive Pull Request!

Thanks for the detailed review!

> When I read it linearly, even if I had the initial context, it was hard to understand some parts as some concepts were defined afterwards. For example, 4.1.2 makes heavy use of concepts in 4.1.3, 4.1.4 and even 4.5. I'm not sure how to fix that. I would recommend creating a section around "Language considerations/overview" that include all or some of 4.5, 4.1.3. and 4.1.4 and put that section early in the document, but it might not play nice with the current document structure that matches model objects or properties.

We could add a subsection about language handling in the introduction - maybe as a separate issue and editorial pull request assuming we agree on this substantive PR? I agree that a separate main section would be hard to fit into the current structure.

> Also, I wonder what we gain by permitting languages that are briefly spoken not to be in the original language list? If all original languages were mandated in the list, you could easily tell all the languages in the document by looking at the document header, and avoid the edge case of a langSrc not listed in the originalLangs list.

I'm uncertain about what the practical use of the original language list will be - perhaps you have specific usages in mind? One of the changes we agreed was to make it optional rather than mandatory, since there's now no semantic that substantively depends on it. If it isn't necessary at all, I don't think we can or should be be too strict about what it contains.

figures/sources/class-diagram.puml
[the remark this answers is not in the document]
nigelmegitt: I've created a definition synonym for Original Languages List too.

examples/event-desc-multiple-descType.xml
```diff
@@ -3,20 +3,20 @@
     <div begin="10s" end="13s">
       <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
       <ttm:desc daptm:descType="plotSignificance">High</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
```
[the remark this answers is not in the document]
nigelmegitt: done in 13deb33

examples/event-desc.xml
```diff
@@ -2,19 +2,19 @@
   <body>
     <div begin="10s" end="13s">
       <ttm:desc>Scene 1</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
```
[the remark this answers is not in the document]
nigelmegitt: done in 13deb33

figures/sources/class-diagram.puml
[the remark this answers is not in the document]
nigelmegitt: done in 13deb33

profiles/dapt-processor-profile.xml
```diff
@@ -85,6 +85,7 @@
     <extension value="required">#agent</extension>
     <extension value="required">#contentProfiles-root</extension>
     <extension value="required">#onScreen</extension>
+    <extension value="required">#originallLang-root</extension>
```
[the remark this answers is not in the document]
nigelmegitt: done in 13deb33

profiles/dapt-content-profile.xml
```diff
@@ -85,13 +85,14 @@
...
     <extension value="required">#scriptType-root</extension>
     <extension value="required">#serialization</extension>
-    <extension value="required">#textLanguageSource</extension>
     <extension value="required">#workflowType-root</extension>
     <extension value="required">#xmlId-div</extension>
     <extension value="required">#xmlLang-root</extension>
     <!-- optional (voluntary) extension support -->
     <extension value="optional">#agent</extension>
     <extension value="optional">#onScreen</extension>
+    <extension value="optional">#originallLang-root</extension>
```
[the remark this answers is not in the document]
nigelmegitt: done in 13deb33

index.html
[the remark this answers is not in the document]
nigelmegitt: done in 13deb33 and 3795d62

nigelmegitt:
In off-thread discussion with @andreastai he expressed a concern that requiring both `daptm:langSrc` and `xml:lang` on all transcriptions of linguistic content could cause unnecessary verbosity.

One possible solution for that would be to make the semantics of inheritance for `daptm:langSrc` match those of `xml:lang` and define that an _empty_ `daptm:langSrc` is equivalent to a _missing_ one, i.e. represents a transcript of content without an inherent language.

This is quite a tempting further improvement to me.

Another possible solution would be to define an explicit attribute for signalling that the source content did or did not have an inherent language. That could also be helpful for tracking when even translations are of content that does not exist directly in the source media.

index.html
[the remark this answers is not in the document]
cconcolato: Thanks, you're right this should be handled with #52

examples/event-desc-multiple-descType.xml
```diff
@@ -3,20 +3,20 @@
...
-      <p xml:lang="fr" daptm:langSrc="translation">
+      <p xml:lang="fr" daptm:langSrc="en">
         <span>Une femme monte à bord d'un petit bateau à voile.</span>
       </p>
     </div>
     <div begin="18s" end="20s">
       <ttm:desc daptm:descType="scene">Scene 1</ttm:desc>
       <ttm:desc daptm:descType="plotSignificance">Low</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
```
cconcolato, line 16: Should we add `xml:lang="en"`?

examples/event-desc.xml
```diff
@@ -2,19 +2,19 @@
...
       </p>
-      <p xml:lang="fr" daptm:langSrc="translation">
+      <p xml:lang="fr" daptm:langSrc="en">
         <span>Une femme monte à bord d'un petit bateau à voile.</span>
       </p>
     </div>
     <div begin="18s" end="20s">
       <ttm:desc>Scene 1</ttm:desc>
-      <p daptm:langSrc="original">
+      <p>
```
cconcolato, line 14: Should we add `xml:lang="en"`?

index.html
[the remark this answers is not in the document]
cconcolato: It seems to me that it would be precisely in the defined term sense here. What confused me is the term 'own'. After reading it again, I'm fine with it.

index.html
```diff
@@ -915,37 +1027,105 @@ <h4>Text</h4>
...
           <h4>Text Language Source</h4>
-          <p>The <dfn>Text Language Source</dfn> property is an annotation indicating whether a <a>Text</a> object is
-            in the same language as the relevant part of the audio's
-            language (original), or if it is a representation in another language (translation):</p>
+          <p>The <dfn>Text Language Source</dfn> property is an annotation indicating the source language of
+            a <a>Text</a> object that represents content with an inherent language.
+            If it differs from the language of the <a>Text</a> object,
+            that implies that it is a representation in another language,
+            referred to as &quot;translation&quot;;
+            conversely if it is the same that implies that the language of the <a>Text</a> is <a>Original</a>.</p>
```
cconcolato, line 1035: This sentence seems duplicative of the sentence at line 1059.

index.html
```diff
@@ -915,37 +1027,105 @@ <h4>Text</h4>
...
+                <li>a description of the scene in the programme video.</li>
               </ol>
             </li>
-            <li><dfn>Translation</dfn> - the <a>Text</a> is a representation of the Original language text in a different language.</li>
+            <li><dfn>Translation</dfn> - the <a>Text</a> is a representation of an <a>Original</a> <a>Text</a> object in a different language.</li>
           </ul>
           <aside class="note">The language of a <a>Text</a> object whose <a>Text Language Source</a> is <a>Original</a>
-            can be different to the document's <a>Primary Language</a>.
+            can be different to the document's <a>Default Language</a>.
+            Similarly it can be absent from the document's <a>Original Languages</a>.
```
cconcolato, line 1053: I wonder if the "Similarly ..." part should not be in its own, separate note. 
The first note is about the case of a source programme using multiple languages. We already have examples in the spec for that (EXAMPLE 11). 
The second note goes into a more specific use case, where that language is not in the original languages list. If not splitting the notes, at least, I would replace "Similarly" with "Furthermore" or "Additionally". I find "Similarly" confusing.

index.html
```diff
@@ -915,37 +1027,105 @@ <h4>Text</h4>
...
-                <li>representative of non-dialogue sound and is in the <a>Primary Language</a> of the <a>DAPT Script</a>;</li>
-                <li>representative of the scene in the original programme video, and is in the <a>Primary Language</a> of the <a>DAPT Script</a>.</li>
+                <li>a transcription of text visible in the programme video, in the same language as that text;</li>
+                <li>representative of non-dialogue sound;</li>
+                <li>a description of the scene in the programme video.</li>
               </ol>
             </li>
-            <li><dfn>Translation</dfn> - the <a>Text</a> is a representation of the Original language text in a different language.</li>
+            <li><dfn>Translation</dfn> - the <a>Text</a> is a representation of an <a>Original</a> <a>Text</a> object in a different language.</li>
           </ul>
```
cconcolato, line 1050: Nitpick: 
- Why is it a bulleted list? I would not use it unless we have an introductory sentence that says something like: "Based on the text language source property, text objects can be of two types:". 
- I would use the same structure as the definition of Original for Translation, that is: "Text is defined as Translation if it is a representation of an Original Text object in a different language"
- Should this (and the NOTE) rather go in the Text object definition?

index.html
```diff
@@ -626,35 +646,115 @@ <h4>Script Type</h4>
...
-            to the language being spoken for the longest duration, or to the language arbitrarily chosen by the author.</p>
-          <p class="note">Text content is marked either as being either in the <a>Original</a> language or
-            as being a <a>Translation</a> independently of the <a>Primary Language</a>, using the <a>Text Language Source</a> property.</p>
-          <aside class="example">An <a>Original Language Transcript</a> <a>script</a> is prepared for a video
+            When multiple languages are used, the <a>Default Language</a> can correspond to the language of the majority of <a>Script Events</a>,
+            to the language being spoken for the longest duration, or to a language arbitrarily chosen by the author.</p>
+          <p class="note">The language of text content is marked either as being either <a>Original</a>  or
+            <a>Translation</a> independently of the <a>Default Language</a>,
+            using the absence or presence of the <a>Text Language Source</a> property,
+            and its value.</p>
```
cconcolato, line 670: This note feels redundant with other parts of the specification and since it does not directly related to "Default Language" I would remove it.

cconcolato:
> We could add a subsection about language handling in the introduction - maybe as a separate issue and editorial pull request assuming we agree on this substantive PR? I agree that a separate main section would be hard to fit into the current structure.

What I would suggest at least is to move the section about default language and original languages before the script type. I think one can understand these 2 sections without understanding script types, and in fact it helps to have them understood before reading script types.

cconcolato:
> I'm uncertain about what the practical use of the original language list will be - perhaps you have specific usages in mind? 

The use case I have in mind is the one I described: knowing all languages used in a programme.

> One of the changes we agreed was to make it optional rather than mandatory, since there's now no semantic that substantively depends on it. 

Generally speaking, I still think there is value in being able to validate DAPT documents with script type specific rules (see #75). For example, when script type is original, then all events shall be in an original language. 

I can agree with the current PR anyway. It's a substantial improvement.

cconcolato:
> In off-thread discussion with @andreastai he expressed a concern that requiring both `daptm:langSrc` and `xml:lang` on all transcriptions of linguistic content could cause unnecessary verbosity.
> 
> One possible solution for that would be to make the semantics of inheritance for `daptm:langSrc` match those of `xml:lang` and define that an _empty_ `daptm:langSrc` is equivalent to a _missing_ one, i.e. represents a transcript of content without an inherent language.
> 
> This is quite a tempting further improvement to me.

I'm not that concerned about verbosity, I prefer readability. The use case for these documents is not transmission over limited networks. That said, I'm not opposed to the suggestion. To make sure I understand, when you say "equivalent" you mean that the semantics would be replaced to say: 
1. if empty, then it means the content does not have an inherent language. 
2. Whether empty or not, when absent, the value is inherited from the parent.

I wonder then if originalLangs and langSrc should not be merged. 

> 
> Another possible solution would be to define an explicit attribute for signalling that the source content did or did not have an inherent language. That could also be helpful for tracking when even translations are of content that does not exist directly in the source media.

Can you draft an example?

index.html
nigelmegitt: Yes, one is describing the data model, the other the syntactic representation. The syntactic representation may change when I change `daptm:langSrc` to be more like `xml:lang` in its inheritance model.

nigelmegitt:
>To make sure I understand, when you say "equivalent" you mean that the semantics would be replaced to say:
>
>  1. if empty, then it means the content does not have an inherent language.
>  2.  Whether empty or not, when absent, the value is inherited from the parent.

Yes, exactly.

> I wonder then if originalLangs and langSrc should not be merged.

I don't understand how that would work.

examples/event-desc-multiple-descType.xml
nigelmegitt: Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630)

examples/event-desc.xml
nigelmegitt: Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630)

index.html
nigelmegitt: No change needed, I think?

index.html
nigelmegitt: Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630) - please check the new wording before resolving this conversation.

index.html
nigelmegitt: Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630)

index.html
nigelmegitt: Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630)

index.html
```diff
@@ -626,35 +646,115 @@ <h4>Script Type</h4>
...
+        </section>
+
+        <section>
+          <h4>Original Languages</h4>
+          <p>
+            The <dfn data-lt="Original languages list">Original Languages</dfn>,
+            also referred to as "Original languages list",
+            is an optional property of a <a>DAPT Script</a>
+            which represents the list of languages used in the original content,
+            and therefore for Text in the <a>Original Language Transcript</a>.
```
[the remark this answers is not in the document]
nigelmegitt: Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630) - please check the new wording before resolving this conversation.

nigelmegitt:
> What I would suggest at least is to move the section about default language and original languages before the script type.

Done in [01d26df](https://github.com/w3c/dapt/pull/179/commits/01d26dfa237f62aa96d98db149b0e3fe825bd630)

nigelmegitt:
Thanks @andreastai - in that case we may be nearly ready to merge this.

One thing I think we have a little vaguely at the moment is the concept of inheritance of `daptm:langSrc` into child `<span>` elements. The text at the moment suggests that it _is_ inherited from a `<p>` into its descendant `<span>`s, but that it can also be set to a different. This means there is an asymmetry, where it is possible to override an inherited `langSrc` with a different one, but not to _clear_ it.

I'm not sure if there are any possible use cases for doing that, but even so, I would like to clarify alongside the related suggestions by @cconcolato in https://github.com/w3c/dapt/pull/179#issuecomment-1828846945 that:
* elements are allowed to have an empty `daptm:langSrc`, i.e. `daptm:langSrc=""`
* `daptm:langSrc=""` implies no inherent language in the source
* `daptm:langSrc=""` is the default value
* when the `daptm:langSrc` attribute is absent from an element, it inherits the value from its parent element.

I think this actually introduces the inheritance that you're worried about @andreastai

> I am bit worried that it would lead to more errors in the application.

Can you be more specific about the error scenarios you're worried about? Possibly the way to address this is to have SHOULD or SHALL statements about the presence of `daptm:langSrc` on `<p>` elements, to reduce the errors, much as we do with `xml:lang`.

nigelmegitt:
Those edits described at https://github.com/w3c/dapt/pull/179#issuecomment-1831523515 now complete as of 490dd02.

index.html
nigelmegitt: This has actually been adjusted now - there is still some duplication, but it is hopefully clearer now that one is about the data model, the other about the XML representation.

nigelmegitt:
@cconcolato @andreastai possibly all the (excellent) discussion points have now been resolved - I would like approval review(s) before merging though, if you have a chance to do a maybe-final review?

nigelmegitt:
No new approvals since the most recent, quite significant changes. Adding to the agenda for Thursday's call.

index.html
```diff
@@ -813,8 +912,8 @@ <h3>Script Event</h3>
             <li>Zero or more <a>Character Identifiers</a> indicating the <a>Characters</a> involved in this <a>Script Event</a>.
               <p class="note">While typically, a <a>Script Event</a> corresponds to one single <a>Character</a>, there are cases where multiple <a>characters</a> can be associated with a <a>Script Event</a>. This is when all <a>Characters</a> speak the same text at the same time.</p>
             </li>
-            <li><p>Zero or more <a>Text</a> objects, each representing either the <a>Original</a> language <a>transcript</a> or
-              <a>Translations</a> of the original language <a>script</a> in other languages.</p>
+            <li><p>Zero or more <a>Text</a> objects, each representing either the <a>Original</a> (untranslated) <a>transcript</a> or
+              <a>Translations</a> in other languages.</p>
```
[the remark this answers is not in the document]
nigelmegitt: I will make a slightly simpler proposal that says the same thing.

index.html
```diff
@@ -613,8 +742,9 @@ <h4>Script Type</h4>
                   They MAY also contain links to audio and mixing instructions for the purpose of producing an audio track incorporating the recordings.</p>
                 <aside class="note">Even though the output of the dubbing workflow is in the same language as those
                 <a>Text</a> objects with the matching language,
-                the <a>Text Language Source</a> of those Text objects remains <a>Translation</a>
-                since they differ from the original language of the dialogue.</aside>
+                the <a>Text Language Source</a> is retained,
+                indicating the source language of those <a>Translation</a> <a>Text</a> objects 
+                since they differ from the untranslated language of the dialogue.</aside>
```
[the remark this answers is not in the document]
nigelmegitt: I think it isn't worded very well, but it is relevant, and it is related to As-Recorded Scripts.

The potential issue is that after a dubbing recording has been made, someone could philophically decide that the As-Recorded Script should be considered a _transcript_ of the dubbed version instead of a translation of the original version. I think we want to discourage that - do you agree?

index.html
```diff
@@ -565,19 +677,31 @@ <h4>Script Type</h4>
...
+                  or as additional context to assist in the adaptation process.
+                  Each <a>Translation</a> <a>Text</a> object indicates its source language
+                  using the <a>Text Language Source</a> property.</aside>
                 <aside class="example">If a programme contains dialogue in English and Hebrew,
                   the French <a>Translated Transcript</a> will contain at least the translation in French of all <a>Script Events</a>.
                   It may still retain text content in Hebrew and English to assist further processing.</aside>
-              </li>
+                <aside class="example">If an <a>audio description</a> <a>Original Language Transcript</a> contains
+                  <a>Original</a> language <a>Script Events</a> that describe in-image text content whose language is not
+                  in the <a>Original languages list</a>, and the desired <a>audio description</a> output needs to be in
```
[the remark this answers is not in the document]
nigelmegitt: Agreed, will delete.

index.html
```diff
@@ -565,19 +677,31 @@ <h4>Script Type</h4>
...
-                  <a>Text Language Source</a> is set to <a>Translation</a>.
-                  They MAY also contain <a>Original</a> language <a>Text</a>.</p>
+                  <a>Text Language Source</a> is set to the source language from which they were translated,
+                  and whose language is set to their actual language,
+                  indicating that they are <a>Translations</a>.
+                  They MAY also contain <a>Original</a> <a>Text</a> objects.</p>
                 <aside class="note">A <a>Script Event</a> in this type of <a>transcript</a> has a 
                   <a>Translation</a> <a>Text</a> object in the <a>Target Recording Language</a>
                   and can also contain an <a>Original</a> <a>Text</a> object, for context,
                   to assist in the adaption process.
```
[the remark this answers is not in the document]
nigelmegitt: Agreed.

index.html
```diff
@@ -523,6 +524,116 @@ <h4>Workflow Type</h4>
...
+              <a>Original languages list</a>.</p>
+            <p>The video image transcript is being prepared for a Danish audience initially,
+              and has its <a>Original Languages</a> set to Danish only
+              by setting <code>daptm:originalLangs="da"</code> on the <code>&lt;tt&gt;</code> element.
+              The <a>Text</a> objects are each marked up with an appropriate language using <code>xml:lang</code>,
+              and, if representing in-image text, a <a>Text Language Source</a> set to
+              the language of that in-image text using <a><code>daptm:langSrc</code></a>.
+              During the video, a Swedish newspaper headline is visible.
+              This is transcribed with both the language and <a>Text Language Source</a> set to Swedish.
+              Those <a>Text</a> objects representing parts of the scene that have no inherent language
```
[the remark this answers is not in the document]
nigelmegitt: I think "Any" would be a better alternative word here. Even though the use of "those" here is correct formal English I can see that it could be confusing or misleading and is a good candidate for using simpler language. Will change it.

index.html
```diff
@@ -423,8 +423,8 @@ <h4>Dubbing Examples</h4>
         </pre>
 
         <p>After translating the text, the document is modified. It includes translation text, and
-        in this case the original text is preserved. The main document language is changed to indicate
-        that the focus is on the translated language:</p>
+        in this case the original text is preserved. The main document's <a>default language</a> is changed to indicate
+        that the focus is on the translated language, but the <a>Original languages list</a> remains the same:</p>
```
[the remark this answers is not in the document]
nigelmegitt: I think we do need to introduce `daptm:langSrc`, and that this example is probably not where the first use _should_ be: it should be in Example 8. I'll add it there and make some references to it there and here.

index.html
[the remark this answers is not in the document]
nigelmegitt: Will propose a wording change, happy to discuss further.

index.html
```diff
@@ -565,19 +677,31 @@ <h4>Script Type</h4>
                 <p>It can be adapted to produce a <a>Pre-Recording Script</a>,
                   and/or used as the basis for a further translation into the <a>Target Recording Language</a>.</p> 
                 <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Translation</a>.
-                  They MAY also contain <a>Original</a> language <a>Text</a>.</p>
+                  <a>Text Language Source</a> is set to the source language from which they were translated,
+                  and whose language is set to their actual language,
+                  indicating that they are <a>Translations</a>.
```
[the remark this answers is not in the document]
nigelmegitt: Will reword, but not sure about having a defined term - what would the term be?

index.html
```diff
@@ -869,7 +968,28 @@ <h3>Script Event</h3>
...
+            </ol>
+          </p>
+          <p><a>Text</a> is defined as <dfn>Translation</dfn> if it is
+            a representation of an <a>Original</a> <a>Text</a> object in a different language.</p>
+          <p>The source language of <a>Translation</a> and, where applicable, <a>Original</a> <a>Text</a> objects
+            is indicated using the <a>Text Language Source</a> property.</p>
+          <aside class="note">The language of an <a>Original</a> <a>Text</a> object
+            can be different to the document's <a>Default Language</a>.</aside>
+          <aside class="note">The language of a <a>Text</a> object whose <a>Text Language Source</a> is <a>Original</a>
+            can be absent from the document's <a>Original Languages</a>.
```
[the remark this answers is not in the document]
nigelmegitt: Will do, but change "may" to "can" so I don't use conformance language in a note.

index.html
```diff
@@ -869,7 +968,28 @@ <h3>Script Event</h3>
...
+            For example a <a>transcript</a> of a short piece of dialogue spoken in a different language to the majority of the dialogue would be annotated
+            with both a different <code>xml:lang</code> value and as being <a>Original</a>,
```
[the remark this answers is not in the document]
nigelmegitt: I'll check for this.

index.html
```diff
@@ -915,37 +1057,117 @@ <h4>Text</h4>
...
+          <p>The <dfn>Text Language Source</dfn> property is an annotation indicating the source language of
+            a <a>Text</a> object, if applicable, or that the source content had no inherent language.</p>
+          <p>If it is not empty and differs from the language of the <a>Text</a> object,
+            that implies that it is a representation in another language,
+            referred to as &quot;translation&quot;;
+            conversely if it is the same that implies that the language of the <a>Text</a> is <a>Original</a>.</p>
+          <p>An empty <a>Text Language Source</a> property on a <a>Text</a> object
+            indicates that the <a>Text</a> represents content without an inherent language,
+            such as untranslated descriptions of a visual scene or
+            captions representing non-dialogue sounds.</p>
```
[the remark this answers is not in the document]
nigelmegitt: I'm happy to make this more structured. The proposal you make here mixes up whether the value is specified or not and the interpretation of the computed value, so it will need some adjustment. There is an if/else for working out the computed value based on what is specified, and then there's an if/else for working out if it's Original or Translation based on that computed value.

index.html
```diff
@@ -915,37 +1057,117 @@ <h4>Text</h4>
...
+            <li>If it is present on an element,
+              the computed value is the specified value.
+            </li>
+            <li>If it is not present on an element,
+              the computed value of the attribute on that element is
+              the computed value of the same attribute on the element's parent,
+              or if the element has no parent it is the default value.
+              <p class="note">The inheritance model of <a><code>daptm:langSrc</code></a> is intended to match
+                the inheritance model of the <code>xml:lang</code> attribute [[XML]].</p>
+            </li>
```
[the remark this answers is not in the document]
nigelmegitt: Nice.

index.html
```diff
@@ -915,37 +1057,117 @@ <h4>Text</h4>
...
+              then it indicates that the <a>Text</a> is a <a>translation</a>,
+              and the computed value is the language from which the <a>Text</a> was translated.</li>
+          </ul>
+          <p class="note">An example of the usage of <a>Text Language Source</a> in a document is present in the <a>Text</a> section.</p>
+          <aside class="example">
+            <table class="simple">
+              <caption>Table enumerating example values of <code>xml:lang</code> and
+                <code>daptm:langSrc</code> for different <a>Original</a> transcript sources and
+                their inherent languages in a document whose <a>Original languages list</a> is
+                English, Arabic.
```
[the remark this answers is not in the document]
nigelmegitt: I don't understand why we would need to add this - the `xml:lang` is already a column in the table, and the Default Language doesn't matter for any of the rows.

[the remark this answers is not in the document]
nigelmegitt: On further thought, this helps to clarify why the content with no inherent language gets an `xml:lang="en"`. Ok.

index.html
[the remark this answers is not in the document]
nigelmegitt: I think this is okay, you may find some edge cases where it could be different if you look hard enough, but I'm not seeing anything that's inconsistent enough to make me want to change it.

nigelmegitt:
Need to fix up the SVG to pass validation

index.html
```diff
@@ -523,6 +531,116 @@ <h4>Workflow Type</h4>
...
+            This language may be one of the <a>Original Languages</a> or a <a>Translation</a> language.
+            When it represents a <a>Translation</a> language, it may be the final language
+            for which a dubbing or audio description <a>script</a> is being prepared,
+            called the <dfn>Target Recording Language</dfn> or it may be an intermediate, or pivot, language
+            used in the workflow.
+          </p>
+
+          <p>The <a>Default Language</a> is represented in a <a>DAPT Document</a> by the following structure and constraints:
+            <ul>
+              <li>the <code>xml:lang</code> attribute MUST be present on the <code>&lt;tt&gt;</code> element and its value MUST NOT be empty.</li>
```
andreastai, line 548: This is an important restriction of TTML2. It needs to be documented in the feature list in the appendix but I am not to sure which feature needs to be commented

index.html
```diff
@@ -523,6 +531,116 @@ <h4>Workflow Type</h4>
...
+          <p>
+            The <dfn data-lt="Original languages list">Original Languages</dfn>,
+            also referred to as "Original languages list",
+            is an optional property of a <a>DAPT Script</a>
+            which represents the languages used in the original content,
+            in other words,
+            the set of languages associated with <a>Text</a> objects in the <a>Original Language Transcript</a>.
+          </p>
+          <p class="note">
+            <a>Text</a> marked as <a>Original</a> can be in a
```
andreastai, line 578: Only minor: text is not `marked` anymore as Original. Text is defined as Original if one of four statements are true. Therefore it might be better to replace `marked` with `defined`.

index.html
nigelmegitt: It would be `#xmlLang-root`.

index.html
nigelmegitt: Oh, @cconcolato wanted consistency in the terminology here too, and seemed to be in favour of using "marked as" in https://github.com/w3c/dapt/pull/179#discussion_r1421218985, and even suggested defining something related to this as a term.

index.html
```diff
@@ -523,6 +531,116 @@ <h4>Workflow Type</h4>
...
+          </p>
+          <p>The <a>Original Languages</a> are represented in a <a>DAPT Document</a> by the following structure and constraints:</p>
+          <ul>
+            <li>the <dfn><code>daptm:originalLangs</code></dfn> attribute
+              SHOULD be present on the <code>&lt;tt&gt;</code> element
+              and if present its value MUST NOT be empty.</li>
+            <li>the value of the attribute MUST be a set of space separated language identifiers,
+              each being a language tag as defined by [[BCP47]].</li>
+          </ul>
+          <!-- BCP47's syntax does not permit white space within each language tag -->
```
andreastai, line 602: There must be a space between `BCP` and `47`, so `BCP 47` instead of `BCP47`. This can be important if people search for `BCP 47`.

nigelmegitt: The system generates a bibliography link in the references section automatically based on this, which is a code from Specref. If they want to follow what the reference points to, they don't need to search for it, they can click on it directly.

The HTML comment is an editorial comment for future potential editors, to remind them why this space separated list is a safe thing to do - I don't think it hugely matters if this has "BCP47" or "BCP 47" but I prefer to have consistency with the reference in the specification.

index.html
andreastai: My understanding of "marked" is that there is additional Markup. But I am fine with any other solution.

index.html
andreastai: For references, it is accepted to use an acronym (I think spaces are not allowed). In another context, I would prefer the correct spelling of the term because this eases manual and automatic search/detection for/of terms (there is also `BCP-47` in the comment of an expression). But as this only a minor issue I would not insist on a change.

index.html
nigelmegitt: I'm happy to make a change here, I'm just not sure _which_ change to make!

andreastai: Somehow I can not access the comment from @cconcolato but if in the document "marked as" is consistently used with the meaning of "has the semantics of" I am also fine with keeping it as is.

index.html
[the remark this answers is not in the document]
cconcolato: It's not really the "definition" part that is important, but the link. It's not obvious how you "mark" something as Original or Translation.

index.html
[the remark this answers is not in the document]
cconcolato: I think I disagree. The "As Recorded Script" is precisely a transcript of what has been recorded in the dubbed audio. Adjustments have been made after the translation from the original version.

index.html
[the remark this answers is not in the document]
cconcolato: I'm not sure what you mean by "mixes up", I just added if/else and bullet points, the text was already there.

index.html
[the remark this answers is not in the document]
cconcolato: I find it confusing to have 3 language attributes on the same element that have the same value: `xml:lang="fr"`, `daptm:langSrc="fr"`, and `daptm:originalLangs="fr"`, when we have only 2 concepts (current language, original language). Maybe we just drop `originalLangs` completely (you can compile its value by scanning the list of langSrc in the content) or maybe we merge langSrc and originalLangs (what I suggested earlier). I need to think about it a bit more.

index.html
```diff
@@ -565,19 +677,31 @@ <h4>Script Type</h4>
                 <p>It can be adapted to produce a <a>Pre-Recording Script</a>,
                   and/or used as the basis for a further translation into the <a>Target Recording Language</a>.</p> 
                 <p><a>Script Events</a> in this type of <a>transcript</a> SHOULD contain <a>Text</a> objects whose
-                  <a>Text Language Source</a> is set to <a>Translation</a>.
-                  They MAY also contain <a>Original</a> language <a>Text</a>.</p>
+                  <a>Text Language Source</a> is set to the source language from which they were translated,
+                  and whose language is set to their actual language,
```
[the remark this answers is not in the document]
cconcolato: I see you rephrased differently. Upon reading again, I wonder if we should not simply remove the part 
> marked as such by having their Text Language Source set to the source language from which they were translated, and their language set to their actual language

If "Translation Text" links to a clear specification of how to signal it there is no need to repeat it here (and elsewhere).

index.html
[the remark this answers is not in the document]
cconcolato: Upon reading the new text, I understand your point. I'm not sure of the value of the note though. Is there value in knowing that something is not a (pure) transcript or is it purely theoretical? I would still delete the note.

index.html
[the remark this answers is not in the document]
cconcolato: Can we avoid using "transcript" in the new wording and simply say: 
> Zero or more Text objects, each being either Original or Translation.

index.html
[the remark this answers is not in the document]
cconcolato: Upon reading your proposed text, I'm fine with it. 

One minor suggestion: I would move the if/Otherwise bullets before the sentence "Text Language Source is an inheritable property.". The If/Otherwise part goes nicely with the first sentence of the section, which I would also end with `:` to emphasize that the bulleted list is providing more precise definition.

index.html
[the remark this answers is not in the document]
cconcolato: I would suggestion using the following:

In "Original Language Transcript":
> Script Events in this type of transcript SHOULD contain Original Text objects and SHOULD NOT contain Translation Text objects.

and:
> EXAMPLE 13
> If a programme contains dialogue in English and Hebrew, the Original Language Transcript will contain some Script Events in English and some in Hebrew, all containing Original Text. None of the Script Events has Translation Text.

In "Translated Transcript":
> Script Events in this type of transcript SHOULD contain Translation Text objects. They MAY also contain Original Text objects.

> EXAMPLE 15
If an audio description Original Language Transcript contains Original language Script Events that describe in-image text content, and the desired audio description output needs to be in a different language, the Text objects can be Translation Text objects. In that case, it is appropriate to mark the transcript as a Translated Transcript. It may also be appropriate, as part of the translation activity, to describe within the translated text the original language of that text, to inform the audience.

In "Pre-recording script":
> NOTE
The Script Type of a DAPT Script cannot necessarily be detected by inspecting the text content of the document. For example, the adaptation of a Translated Transcript into a Pre-recording Script can consist in replacing some words in the text content of a Script Event without altering the rest of the document. In either case, Translation Text objects have their Text Language Source properties set to the source language from which they were translated.

> NOTE
The Original Text objects in Audio Description Script Events do not have a Text Language Source property if they represent visual elements of the scene that do not have an inherent language, but are required to have a Text Language Source if they do represent visual elements with an inherent language, such as in-image text. If audio description scripts are translated, their translations would be represented by Translation Text objects.

index.html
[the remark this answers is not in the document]
cconcolato: I'm resolving this conversation because I made a simplifying proposal in https://github.com/w3c/dapt/pull/179#discussion_r1428367490 which I think resolves it.

index.html
cconcolato: I made a proposal in https://github.com/w3c/dapt/pull/179#discussion_r1428367490. Essentially, I suggest:
- not defining any new term (e.g. marked as)
- stop repeating the semantics of Original or Translation text
- using the term "marked as" and directly refer to Original Text objects or Translation Text objects.

andreastai: > not defining any new term (e.g. marked as)
> using the term "marked as" and directly refer to Original Text objects or Translation Text objects.

@cconcolato Do you mean "**not**  using the term 'marked as'..."? Otherwise fine with the proposals in the three bullet points.

andreastai:
@nigelmegitt Apologies for the late reply to your comment in https://github.com/w3c/dapt/pulls#issuecomment-1831523515. 

After reading through the current PR I would be in favor for allowing `@daptm:langSrc` only on `<p>` and `<span>` elements (so, only to the elements where it applies). Furthermore I propose to make the the attribute mandatory on `<p>` elements (one of the options you propose in your comments).

Reasoning: The potential errors scenarios are:

* Users may oversee inherited values and take only the values of `@daptm:langSrc`  on `<p>` and `<span>` elements to conclude if the text objects are `Original` or `Translation`. 

* The default value of `@daptm:langSrc` applies only to a specific use case and not the majority of the use cases. Therefore users may be easily imply unwanted semantics if they do not specify the attribute.

Although you mention these potential errors in the informative note it may better to reduce them by being more restrictive. This would also be better for the requirement of `readability` that @cconcolato mentions in https://github.com/w3c/dapt/pulls#issuecomment-1828846945.

For "readability" it may also be better to make `xm:lang` on text objects mandatory.

index.html
```diff
@@ -564,20 +683,23 @@ <h4>Script Type</h4>
...
                   the French <a>Translated Transcript</a> will contain at least the translation in French of all <a>Script Events</a>.
                   It may still retain text content in Hebrew and English to assist further processing.</aside>
-              </li>
+                <aside class="example">If an <a>audio description</a> <a>Original Language Transcript</a> contains
+                  <a>Original</a> language <a>Script Events</a> that describe in-image text content,
+                  and the desired <a>audio description</a> output needs to be in
+                  a different language, the <a>Text</a> objects can be translated and marked as <a>Translation</a>.
+                  In that case, it is appropriate to mark the transcript as a <a>Translated Transcript</a>.
+                  It may also be appropriate, as part of the translation activity, to describe within the translated text
+                  the original language of that text, to inform the audience.
```
andreastai, line 700: It may be good to illustrate this case with an example. For me the meaning of `describe within the translated text the original language of that text` war not clear.

index.html
```diff
@@ -611,10 +738,10 @@ <h4>Script Type</h4>
                   and MAY also contain <a>Original</a> <a>Text</a> objects from the <a>Original Language Transcript</a>
                   or <a>Translation</a> <a>Text</a> objects in other languages for context and quality verification.
                   They MAY also contain links to audio and mixing instructions for the purpose of producing an audio track incorporating the recordings.</p>
-                <aside class="note">Even though the output of the dubbing workflow is in the same language as those
-                <a>Text</a> objects with the matching language,
-                the <a>Text Language Source</a> of those Text objects remains <a>Translation</a>
-                since they differ from the original language of the dialogue.</aside>
+                <aside class="note"><a>As-recorded Scripts</a> are not considered to be transcripts
+                  even though they are adjusted to match the recorded output of the dubbing workflow.
```
andreastai, line 742: I propose to delete the addition  "As-recorded Scripts</a> are not considered to be transcripts even though they are adjusted to match the recorded output of the dubbing workflow." It would make the note clearer if it just contains the second sentennce.

index.html
```diff
@@ -869,7 +965,31 @@ <h3>Script Event</h3>
...
-         <p>The <dfn>Text</dfn> object contains text content typically in a single language. This language may be the <a>Original</a> language or a <a>Translation</a> language.</p>
+          <p>The <dfn>Text</dfn> object contains text content typically in a single language.
+            This language may be the <a>Original</a> language or a <a>Translation</a> language.</p>
+          <p><a>Text</a> is defined as <dfn>Original</dfn> if it is any of:
+            <ol>
+              <li>the same language as the dialogue that it represents in the original programme audio;</li>
+              <li>a transcription of text visible in the programme video, in the same language as that text;</li>
+              <li>representative of non-dialogue sound;</li>
+              <li>a description of the scene in the programme video.</li>
+            </ol>
```
andreastai, line 976: The second two bullet points may need to be changed to

* **untranslated** description of non-dialogue sound
* an **untranslated** description of the scene in the programme video

index.html
```diff
@@ -885,27 +1005,49 @@ <h4>Text</h4>
...
+                source had an inherent language.
+                <aside class="note">
+                  <p>If a <code>&lt;p&gt;</code> element omits the <a><code>daptm:langSrc</code></a> attribute
+                    then its computed value is derived by inheritance from its parent element,
+                    and so forth up to the root <code>&lt;tt&gt;</code> element.</p>
+                  <p>In <a>scripts</a> that have very little variation in source language,
+                    for example audio description scripts,
+                    the <a><code>daptm:langSrc</code></a> attribute can be set on the root element
+                    and omitted from <code>&lt;p&gt;</code> elements except where its value
+                    differs.
```
andreastai, line 1019: I wonder if audio description scripts are the best example for this use case. In most cases audio description scripts would be descriptions of scenes in the programme audio where the source has no inherent language. Given the current PR text it would not be needed to specify `@daptm:langSrc`. 

(But see also proposal to disallow the attribute on `<tt>`, `<body>` and `<div>` in my comment https://github.com/w3c/dapt/pull/179#issuecomment-1859142948)

index.html
```diff
@@ -611,10 +738,10 @@ <h4>Script Type</h4>
...
                   or <a>Translation</a> <a>Text</a> objects in other languages for context and quality verification.
                   They MAY also contain links to audio and mixing instructions for the purpose of producing an audio track incorporating the recordings.</p>
-                <aside class="note">Even though the output of the dubbing workflow is in the same language as those
-                <a>Text</a> objects with the matching language,
-                the <a>Text Language Source</a> of those Text objects remains <a>Translation</a>
-                since they differ from the original language of the dialogue.</aside>
+                <aside class="note"><a>As-recorded Scripts</a> are not considered to be transcripts
+                  even though they are adjusted to match the recorded output of the dubbing workflow.
+                  The <a>Translation</a> <a>Text</a> objects within them retain their <a>Text Language Source</a>,
+                  so that the source language from which they were translated remains available.</aside>
```
nigelmegitt, line 744: Something like:

```suggestion
                <aside class="note">The <a>Translation</a> <a>Text</a> objects within <a>As-recorded Scripts</a> retain their <a>Text Language Source</a>,
                  so that the source language from which they were translated remains available.</aside>
```

?

index.html
nigelmegitt: Right, can just delete the "for example audio description scripts," line.

nigelmegitt:
> @nigelmegitt Apologies for the late reply to your comment in https://github.com/w3c/dapt/pulls#issuecomment-1831523515.
> 
> After reading through the current PR I would be in favor for allowing `@daptm:langSrc` only on `<p>` and `<span>` elements (so, only to the elements where it applies). Furthermore I propose to make the the attribute mandatory on `<p>` elements (one of the options you propose in your comments).
> 
> Reasoning: The potential errors scenarios are:
> 
>     * Users may oversee inherited values and take only the values of `@daptm:langSrc`  on `<p>` and `<span>` elements to conclude if the text objects are `Original` or `Translation`.
> 
>     * The default value of `@daptm:langSrc` applies only to a specific use case and not the majority of the use cases. Therefore users may be easily imply unwanted semantics if they do not specify the attribute.
> 
> 
> Although you mention these potential errors in the informative note it may better to reduce them by being more restrictive. This would also be better for the requirement of `readability` that @cconcolato mentions in https://github.com/w3c/dapt/pulls#issuecomment-1828846945.
> 
> For "readability" it may also be better to make `xm:lang` on text objects mandatory.

On reflection I don't think this is much about 'users', practically, unless someone is making a very naive authoring implementation, or editing manually. Instead, the constituency that would be most worried about this is probably implementers.

The advantage of allowing inheritance of `daptm:langSrc` and `xml:lang` is to reduce document size and processor time when parsing the document. In a distribution environment and playback those matter to some extent; in an editing and exchange environment, they are not so important.

You've described the disadvantages quite clearly: in this case, inheritance is a vector for introducing unexpected behaviour or bugs if implementers don't deal with this.

My own view is swayed slightly more in favour of reduced duplication and to allow more inheritance, because I _would_ like to see this format adopted in players eventually, for AD.

But I would certainly like to hear any input from implementers. Also, maybe the notes are targeted wrongly and should be more implementation-focused. At the moment it describes mitigations that authors can take, but realistically, authors are unlikely to be making a choice here. I'd be happy to try to draft a warning directly to implementers to point out that when changing `daptm:langSrc` or `xml:lang` on an element they need to check down the tree and be if appropriate specify the value on descendant elements so that their meaning does not change unintentionally.

index.html
cconcolato: @andreastai I modified my comment. I did mean "NOT using".

index.html
cconcolato: Is it missing "can" in "can retain"?

andreastai: I think Translation objects must have a computed value of `daptm:langSrc` other than the empty string. `can` would make it a `MAY` constraint.

index.html
nigelmegitt: I'm going to mark this conversation as resolved for tidiness - @andreastai please unresolve if you think that action is needed.

index.html
nigelmegitt: Done in 10b1a0e

index.html
nigelmegitt: Done in 10b1a0e

index.html
nigelmegitt: Agree with @andreastai we should not make this appear to be optional.

nigelmegitt: I was tempted to separate this into two notes, one for each point, but on close reading I don't think it matters whether As-Recorded scripts are transcripts or scripts - nothing in the spec seems to depend on this distinction. So I'll delete that part.

nigelmegitt: Done in 10b1a0e

index.html
nigelmegitt: Done in 10b1a0e

index.html
nigelmegitt: Done in 74e4017

index.html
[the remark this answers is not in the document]
nigelmegitt: Yes, I agree, it is not clear whether this `daptm:originalLangs` concept has general applicability and is overall helpful. It was introduced as part of the work for this PR and later iterations mean it is not actually needed. It is not something requested in the issue threads that generated this PR. I could imagine that we might bring it back later if requested, but for now, I will delete it.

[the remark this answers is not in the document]
nigelmegitt: Removed Original Languages in 1693603.

index.html
```diff
@@ -869,7 +893,24 @@ <h3>Script Event</h3>
...
+          <p><a>Text</a> is defined as <dfn>Original</dfn> if it is any of:
+            <ol>
+              <li>the same language as the dialogue that it represents in the original programme audio;</li>
+              <li>a transcription of text visible in the programme video, in the same language as that text;</li>
+              <li>an untranslated representative of non-dialogue sound;</li>
+              <li>an untranslated description of the scene in the programme video.</li>
+            </ol>
+          </p>
+          <aside class="note">The language of an <a>Original</a> <a>Text</a> object
+            can be different to the document's <a>Default Language</a>.</aside>
```
nigelmegitt, line 907: Although this _defines_ Original and Translation, it doesn't say how any given Text object can be _identified_ as either Original or Translation.

nigelmegitt: Addressed in d58c992

index.html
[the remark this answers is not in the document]
nigelmegitt: Done in fa7ca74, though the last proposal was slightly overtaken by other changes. However the important part seemed to be the final sentence, for which I made the equivalent change.

nigelmegitt:
> After reading through the current PR I would be in favor for allowing `@daptm:langSrc` only on `<p>` and `<span>` elements (so, only to the elements where it applies).

@andreastai In 1693603 I changed it so that `daptm:langSrc` can only be specified on `tt`, `p` and `span` which matches the data model diagram.

nigelmegitt:
> Also, maybe the notes are targeted wrongly and should be more implementation-focused. At the moment it describes mitigations that authors can take, but realistically, authors are unlikely to be making a choice here. I'd be happy to try to draft a warning directly to implementers to point out that when changing `daptm:langSrc` or `xml:lang` on an element they need to check down the tree and be if appropriate specify the value on descendant elements so that their meaning does not change unintentionally.

Warnings improved to mention implementers as well as authors in 106b951.

nigelmegitt:
@cconcolato @andreastai thank you both for persevering with your reviews on this. I think all the comments have now been addressed, if you'd like to check, and mark any outstanding conversation threads as "resolved" and add positive reviews if you're happy to.

The one conversation where we have maybe not arrived at the same conclusion is whether or not to permit `daptm:langSrc` on `<tt>`. I think I've gone as far as I'd be happy with on this for now - @andreastai are you able to accept it as is?

index.html
cconcolato: FYI, there is a sentence that uses "mark" (not marked), but it is fine to leave it as is:
>  The combination of the xml:lang and daptm:langSrc attributes are used to mark the text as being original or translated

index.html
```diff
@@ -416,15 +423,21 @@ <h4>Audio Description Examples</h4>
         <h4>Dubbing Examples</h4>
         <p>From the basic structure of <a href="#example-1"></a>, a transcription of the audio programme produces an original language dubbing <a>script</a>,
```
cconcolato, line 424: minor editorial: the hyperlink is on the word script. It does not feel right because a transcription would be a transcript. I suggest either putting the hyperlink around "dubbing script" or changing script to transcript.

index.html
```diff
@@ -367,7 +370,11 @@ <h4>Basic document structure</h4>
         <p>The following examples will demonstrate different uses in dubbing and audio description workflows.</p>
 
         <h4>Audio Description Examples</h4>
-        <p>When descriptions are added this becomes a <a>Pre-Recording Script</a>:</p>
+        <p>When descriptions are added this becomes a <a>Pre-Recording Script</a>.
+          Note that in this case, to reflect that most of the audio description content
+          transcribes the video image where there is no inherent language,
+          the <a>Text Language Source</a>, represented by the <a><code>daptm:langSrc</code></a> attribute,
+          is set to the empty string at the top level of the document:</p>
```
cconcolato, line 377: Is it worth mentioning that it could be omitted because the default value is the empty string?

index.html
```diff
@@ -416,15 +423,21 @@ <h4>Audio Description Examples</h4>
...
           data-include-format="text">
         </pre>
 
         <p>After translating the text, the document is modified. It includes translation text, and
-        in this case the original text is preserved. The main document language is changed to indicate
-        that the focus is on the translated language:</p>
+        in this case the original text is preserved. The main document's <a>default language</a> is changed to indicate
+        that the focus is on the translated language.
+        The combination of the <code>xml:lang</code> and <a><code>daptm:langSrc</code></a> attributes are used
+        to mark the text as being original or translated:</p>
```
cconcolato, line 440: I note that Example 9 and 10 are not using inheritance of xml:lang nor daptm:langSrc. This is fine. I just wonder if we should say, the xml:lang and daptm:langSrc have been duplicated here for readability but could be omitted in some places following the inheritance model.

index.html
```diff
@@ -587,20 +637,25 @@ <h4>Script Type</h4>
...
                 
-                <aside class="note">The <a>Text</a> objects in Audio Description <a>Script Events</a> are each expected to have
-                  a <a>Text Language Source</a> property of <a>Original</a> if their language is the <a>Primary Language</a>
-                  or, when representing in-image text,
-                  if they are in the same language as that text.
+                <aside class="note">The <a>Original</a> <a>Text</a> objects in Audio Description <a>Script Events</a>
+                  have an empty <a>Text Language Source</a> property
+                  if they represent visual elements of the scene that do not have an inherent language.
+                  Otherwise if they do represent visual elements with an inherent language, such as in-image text,
+                  they are required to have a <a>Text Language Source</a> that specifies a language.
```
cconcolato, line 655: Can we use "required" in a NOTE?

index.html
```diff
@@ -869,7 +890,27 @@ <h3>Script Event</h3>
...
+          <aside class="note">The language of an <a>Original</a> <a>Text</a> object
+            can be different to the document's <a>Default Language</a>.</aside>
+          <p><a>Text</a> is defined as <dfn>Translation</dfn> if it is
+            a representation of an <a>Original</a> <a>Text</a> object in a different language.</p>
+          <p><a>Text</a> can be <em>identified</em> as being <a>Original</a> or <a>Translation</a>
+            by inspecting its language and its <a>Text Language Source</a> together,
+            according to the semantics defined in <a>Text Language Source</a>.</p>
+          <p>The source language of <a>Translation</a> <a>Text</a> objects and, where applicable,
+            <a>Original</a> <a>Text</a> objects
+            is indicated using the <a>Text Language Source</a> property.</p>
```
cconcolato, line 912: It feels to me that the 2 sentences are saying the same thing. Maybe keep only one?

index.html
```diff
@@ -915,37 +989,124 @@ <h4>Text</h4>
...
             </pre>
           </div>
+          <ul>
+            <li>The value MUST be an empty string or a language identifier as defined by [[BCP47]].</li>
+            <li>The default value is the empty string.</li>
+            <li>It <em>applies</em> to <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code> elements.</li>
+            <li>It MAY be specified on the following elements: <code>&lt;tt&gt;</code>, <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code>.</li>
+            <li>If it is present on an element,
+              the computed value is the specified value.
+            </li>
```
cconcolato, line 1029: That sentence is duplicated below:
```suggestion
```

index.html
```diff
@@ -915,37 +989,124 @@ <h4>Text</h4>
...
+: ""                    # default
+
+&lt;language-identifier&gt;   # valid BCP-47 language tag
             </pre>
           </div>
+          <ul>
+            <li>The value MUST be an empty string or a language identifier as defined by [[BCP47]].</li>
+            <li>The default value is the empty string.</li>
+            <li>It <em>applies</em> to <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code> elements.</li>
+            <li>It MAY be specified on the following elements: <code>&lt;tt&gt;</code>, <code>&lt;p&gt;</code> and <code>&lt;span&gt;</code>.</li>
```
cconcolato, line 1026: I'm still wondering about `div` and `body`.
On the one hand, I understand your argument about matching the data model. On the other hand, aligning with `xml:lang` is pleasant.

figures/sources/class-diagram.puml
```diff
@@ -15,7 +15,8 @@ skinparam MinClassWidth 220
 Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
     Workflow Type [[[#workflow-type]]]
     Script Type [[[#script-type]]]
-    Primary Language [[[#primary-language]]]
+    Default Language [[[#default-language]]]
```
cconcolato, line 18: Why did we change from Primary to Default? I did prefer Primary but won't insist if you still prefer "Default"

cconcolato (approved):
@nigelmegitt Thank you for the work done! It reads great to me now. I still have some minor changes to suggest but they are not blocking, so I'm approving this PR!

index.html
nigelmegitt: I don't think they are saying the same thing. The first is about identifying Original vs Translation, the second is about what the source language was.

figures/sources/class-diagram.puml
nigelmegitt: My concern is that Primary has connotations that could lead to misunderstanding. Primary (also the alternative "Initial") suggests some sort of time or precedence ordering, as though the language specified here is more important, or the starting language for transcription etc. whereas Default is more neutral in that respect, and feels more specific to this particular instance of the document.

index.html
nigelmegitt: Me too - I'm slightly uncomfortable missing `div` and `body` out, but I'm also struggling to understand the use case for including them, other than completeness. This is my "compromise" position between @andreastai's "only on `p` and `span`, and require it" position and my "provide a scheme where omitting it makes sense, especially for audio description". If we find a need to add `div` and `body` back in later, we could do that.

index.html
nigelmegitt: Yes we can. It is not considered "conformance language" as per this text in §6 Conformance:

>The key words MAY, MUST, MUST NOT, SHOULD, and SHOULD NOT in this document are to be interpreted as described in BCP 14 [RFC2119] [RFC8174] when, and only when, they appear in all capitals, as shown here.

index.html
nigelmegitt: Will change to transcript, and fix "dialogs" below, for consistency, because we use "dialogue" everywhere else.

index.html
nigelmegitt: Done in a1c80d4

index.html
nigelmegitt: Done in a1c80d4

index.html
nigelmegitt: Done in a1c80d4

index.html
nigelmegitt: Done in a1c80d4

index.html
```diff
@@ -936,24 +936,36 @@ <h4>Text</h4>
                   <p>In <a>scripts</a> that have very little variation in source language,
                     the <a><code>daptm:langSrc</code></a> attribute can be set on the root element
                     and omitted from <code>&lt;p&gt;</code> elements except where its value
-                    differs.
-                    Care should be taken if using this approach, especially when moving between
+                    differs.</p>
```
andreastai, line 946: It may be good to add that if there is no `daptm:langSrc` is set on `<tt>` and `<p>` (e.g. it is not used at all in the document), all text objects signal that they have no inherent language.

index.html
andreastai: Thanks, @nigelmegitt for the effort. I think this edit is an important improvement. You only need to check in two places to get the value of the source language.

andreastai (approved):
Thanks for all the effort @nigelmegitt I went through all the changes to address my comments and all looks good to me : ) I had just one minor comment that is not blocking and could also be applied after merging the PR.

figures/sources/class-diagram.puml
cconcolato: I do think the language specified here is more important. The note even says "a language arbitrarily chosen by the author".
My concern with "default" is that it is an overloaded term. We use "default value" in plenty of places to refer to the value when not specified.
Could we just say "Language"?

nigelmegitt: I'm going to raise this as a separate follow-on issue.

index.html
nigelmegitt: Done in aad01da

figures/sources/class-diagram.puml
nigelmegitt: Raised as #204

PR #181, opened 2023-08-01
Split Script Event Description out as a separate object

### LINKED ISSUE #174: Script Event Description should be an object
Script Event Description should be represented as an object. With Script Event Description Type (currently only specified in the TTML syntax as the attribute `daptm:descType`) it has its own property and it can also be associated with Script Event multiple times.

nigelmegitt:
Closes #174.

Clarifies the cardinality of Script Event Description.

Updates the data model diagram (in a way that is non-ideal but hard to fix automatically: it may be worth replacing the PlantUML-generated diagram with a more manually constructed one prior to final publication), and defines Description Type more explicitly.

Also clarify that there is no uniqueness constraint for `daptm:descType` and that `xml:lang` can be used to label the language of the contents of a `ttm:desc`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/181.html" title="Last updated on Oct 9, 2023, 9:23 AM UTC (006c394)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/181/09b23e7...006c394.html" title="Last updated on Oct 9, 2023, 9:23 AM UTC (006c394)">Diff</a>


nigelmegitt:
Raised #182 regarding cleaning up the PlantUML-generated SVG diagrams.

nigelmegitt:
pinging @cconcolato @andreastai for a review please

cconcolato (approved):
Should the model be explicit about the language property?

nigelmegitt:
> Should the model be explicit about the language property?

Not sure I understand - are you suggesting that we should require the presence of `xml:lang` on `ttm:desc` elements?

nigelmegitt:
Assuming we merge #185 the PlantUML diagram will need to change again to accommodate this. Or we could merge this first and then rebase #185. Either way, some redesign will be needed, which I can pick up.

cconcolato:
> Not sure I understand - are you suggesting that we should require the presence of `xml:lang` on `ttm:desc` elements?

I wasn't clear. When I read the sentence https://github.com/w3c/dapt/pull/181/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R984, I thought maybe we should have an `(optional) language` line in the UML diagram.

nigelmegitt:
Oh sure, makes sense.

nigelmegitt:
Have added optional Language to the script description in the UML model. The model in this PR originally looked pretty bad, and the one in #185 is much better, from a layout perspective, but omits Style. To save editorial time, I've slightly naughtily assumed that we will approve and merge #185, and gone ahead with the UML model from that pull request, omitting Style, and added the ScriptDescription object in to that one.

If we don't merge #185 then we'll need to put the Style back.

andreastai (commented):
Thanks, @nigelmegitt Overall the changes look good to me. Some minor observations:

The information about the Description Type is now represented two times in the Document:

> The [Description Type](https://github.com/w3c/dapt/pull/181/files#dfn-description-type) is represented in a [DAPT Document](https://github.com/w3c/dapt/pull/181/files#dfn-dapt-document) by a daptm:descType attribute on the <ttm:desc> element.

and 

> The <ttm:desc> element MAY have a daptm:descType attribute specified to indicate the [Description Type](https://github.com/w3c/dapt/pull/181/files#dfn-description-type).

In any case, the info about the `Description Type` should go after Example 20.

The information that the attribute `@xml:lang` is used to represent the optional language of the Event Description should be in the normative text with a MAY keyword rather than in a note. 

A note may also be added if the language of the Event Description is the Primary Language of the DAPT script in case `@xml:lang` is omited on `<ttm:desc>`.

nigelmegitt:
Thank you @andreastai for the great review feedback. I've reworked the section to address the points. Just one thing I disagree with: if `xml:lang` is omitted it inherits its language from the Script Event, which in principle might not be the Primary Language of the document.

Hopefully the order is now clearer and there is no unnecessary duplication.

andreastai:
> Thank you @andreastai for the great review feedback. I've reworked the section to address the points. 

Yes, thanks @nigelmegitt! The edits look good to me.

> Just one thing I disagree with: if `xml:lang` is omitted it inherits its language from the Script Event, which in principle might not be the Primary Language of the document.

As I understand the Script Event object does not have a language property and the language is always specified on the Text object.

nigelmegitt:
> As I understand the Script Event object does not have a language property and the language is always specified on the Text object.

Oh I see. Formally, Script Events do have a computed language regardless of whether they have an `xml:lang` attribute, and that might be inherited from the document level.

My view here is that we should go with the text in this pull request and if anyone asks about this subtlety we can consider amending or adding extra explanation.

andreastai:
> Oh I see. Formally, Script Events do have a computed language regardless of whether they have an xml:lang attribute, and that might be inherited from the document level.

> My view here is that we should go with the text in this pull request and if anyone asks about this subtlety we can consider amending or adding extra explanation.

Thanks, @nigelmegitt I think we should be clear about how language information is applied and inherited. From the class model (at least from what I can see) there is no other way to inherit the language information other than from the document level.

nigelmegitt:
> I think we should be clear about how language information is applied and inherited.

Raised as #192.

PR #183, opened 2023-09-28
Redraft opening section of §2.1

### LINKED ISSUE #175: Consider removal of distinction between script and transcript in chapter 2.1
The distinction between script and transcript in [chapter 2.1](https://w3c.github.io/dapt/#transcripts-and-scripts) is confusing. While script is used here for differentiation, the term 'script,' in combination with the type of domain or workflow, is used as a superset. Thus, a DAPT script or a Dubbing script can be both a script and a transcript. It is worth considering whether the differentiation in 2.1 is redundant since the ScriptType values already sufficiently differentiate various script types.

nigelmegitt:
Closes #175 as per https://github.com/w3c/dapt/issues/175#issuecomment-1715823202 .

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/183.html" title="Last updated on Oct 12, 2023, 4:21 PM UTC (7d8d37d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/183/04b0b5d...7d8d37d.html" title="Last updated on Oct 12, 2023, 4:21 PM UTC (7d8d37d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -272,17 +272,17 @@ table.coldividers td + td { border-left:1px solid gray; }
           <li>The <code>&lt;tt&gt;</code> root element in the namespace <code>http://www.w3.org/ns/ttml</code> indicates that this is a TTML document
           and the <a><code>ttp:contentProfiles</code></a> attribute indicates that it adheres to the DAPT <a>content profile</a> defined in this specification.</li>
           <li>The <code>daptm:workflowType</code> attribute indicates the type of workflow.</li>
-          <li>The <code>daptm:scriptType</code> attribute indicates the type of script
-          but in this empty example, it is not relevant, as only the structure of the document is shown.</li>
+          <li>The <code>daptm:scriptType</code> attribute indicates the type of <a>transcript</a> or <a>script</a>
+          but in this empty example, it is not relevant, since only the structure of the document is shown.</li>
         </ul>
         <p>The structure is applicable to all types of <a>DAPT scripts</a>, dubbing or <a>audio description</a>.</p>
         <pre class="example"
           data-include="examples/intro-top-level.xml"
           data-include-format="text">
         </pre>
-        <p>The following examples correspond to the timed text <a>scripts</a> produced
+        <p>The following examples correspond to the timed text <a>transcripts</a> and <a>scripts</a> produced
           at each stage of the workflow described in [[DAPT-REQS]].</p>
-        <p>The first example shows a <a>script</a> where timed opportunities for descriptions
+        <p>The first example shows an early stage <a>transcript</a> in which timed opportunities for descriptions
           or transcriptions have been identified but no text has been written:</p>
         <pre class="example"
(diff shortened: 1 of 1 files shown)


andreastai (commented):
Thanks @nigelmegitt!

The conclusion in our last discussion (documented in github comment) as as follows:

> MattS: I definitely think they are not used interchangeable. But "script" has a broader range of meanings than "transcript".

> Andreas: Would you say a "transcript" is not a "script"?

> MattS: I would say a "transcript" is a kind of "script".
> MattS: From a workflow point of view a script is forward looking, but linguistically a transcript is a form of script.

> Cyril: How do we resolve this issue? Is modifying the first paragraph to say that a transcript is a special type of script - would that help?

> Andreas: That would be clearer to me at least, but then you have to rewrite the whole paragraph. Currently they seem to be mutually exclusive.

In your edit you make DAPT script a superset of `transcript` and `script` but you keep the distinction of `transcript` and `script`. From my understanding, we concluded that `transcript` is a subset of `script`.

nigelmegitt:
Thanks @andreastai I thought I'd done that, but maybe I'm still missing the point. I just pushed a small change to clarify that in general usage a transcript is a type of script. Does that work?

andreastai:
> Thanks @andreastai I thought I'd done that, but maybe I'm still missing the point. I just pushed a small change to clarify that in general usage a transcript is a type of script. Does that work?

Thanks @nigelmegitt for the addition to clarify the intent. My understanding was that we follow the general usage and do not make the below distinction between transcripts and scripts.

> a script is a text representation of the intended content of media prior to its creation, for example, to guide an actor in recording an audio track.

With this usage, a transcript is not a script. It is now explained that this is done on purpose but I would argue you do not need this different use of `script`. We (nearly) always use the term script in the more generic understanding e.g. in compounds like `dubbing scripts` or in DAPT vocabulary like `script event`.

But if nobody else has an issue with this, I am fine with leaving it like it is.

nigelmegitt:
Added to the agenda for tomorrow's call, to allow us to check in on this and discover if anyone else thinks further change is needed.

CC @mattsimpson3 @cconcolato who were involved in the discussion, especially because Cyril has sent his regrets for the call - comments on this thread would be very useful too as always.

mattsimpson3 (approved):
All good for me!

nigelmegitt:
@andreastai I think I've removed all unwanted uses of "script" as a standalone term, which were all in the Introduction. I did a search through the rest of the document, both in rendered form and in the source, and didn't find any others.

andreastai:
> @andreastai I think I've removed all unwanted uses of "script" as a standalone term, which were all in the Introduction. I did a search through the rest of the document, both in rendered form and in the source, and didn't find any others.

Thanks, @nigelmegitt, I also went through all uses of the term "script" and the use is now consistent.

PR #185, opened 2023-10-04
Remove styles

### LINKED ISSUE #124: Reconsidering the role of character styles
In my opinion, the use of character styles in the specification has limited practical use. Here are a few reasons why:

- A character style is tied to one character and cannot be applied to a group of characters with similar requirements.

- The application of a character style to a script event or text must be done explicitly, and there is no automatic application based on the character referenced through the `@ttm:agent` attribute.

- The only extra information carried by the application of a character style is that it was authored to be applied to a specific character, which may not justify the introduction of this concept.

Currently, the same effect as a character style could be achieved by using a "regular" style, which would allow for the reuse of style objects.

One possible reason for the current design may be to allow for multiple character styles to be applied to a single character. An example use case where this could be helpful is the application of one character style to the original language text and another character style to the translated text associated with the same character. If this is a use case, providing a code example would help to illustrate this scenario more clearly.

Depending on the supported workflow, it could be useful to allow a many-to-many relationship instead of the current one-to-many relationship. This would allow for one style to be applied to multiple characters. In this case, it may be beneficial to reverse the reference direction so that a character object references zero or many styles instead of a style object referencing one character.

### LINKED ISSUE #162: Can Script Events be styled?
This is true, but begs the question: why do we not also say that Script Events can be styled, since they are represented by `<div>`s that do support styling?

_Originally posted by @nigelmegitt in https://github.com/w3c/dapt/pull/153#discussion_r1234258879_

cconcolato:
Close #124
Close #162

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/185.html" title="Last updated on Oct 13, 2023, 1:07 PM UTC (b16db46)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/185/05e829f...b16db46.html" title="Last updated on Oct 13, 2023, 1:07 PM UTC (b16db46)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -630,13 +630,9 @@ table.coldividers td + td { border-left:1px solid gray; }
           <ul>
             <li>a mandatory <dfn data-lt="Character Identifier">Identifier</dfn>
             which is a unique identifier used to reference the character from elsewhere in the document,
-            for example to indicate when a <a>Character</a> participates in a <a>Script Event</a>,
-            or to link a <a>Character Style</a> to its <a>Character</a>.</li>
+            for example to indicate when a <a>Character</a> participates in a <a>Script Event</a>.</li>
             <li>a mandatory <dfn data-lt="Character Name">Name</dfn> which is the name of the <a>Character</a> in the programme</li>
-            <li>an optional <dfn data-lt="Character Talent Name">Talent Name</dfn>, which is the name of the actor speaking dialogue for this <a>Character</a></li>
-            <li>zero or more <dfn data-lt="Character Style">Character Style</dfn> objects
-              which can be applied to control the visual appearance of <a>Script Events</a> spoken by the <a>Character</a>,
-              for example during recording by an actor or when transforming the <a>script</a> into subtitles.</li>
+            <li>an optional <dfn data-lt="Character Talent Name">Talent Name</dfn>, which is the name of the actor speaking dialogue for this <a>Character</a>.</li>
           </ul>
         </p>
         <p>A <a>Character</a> is represented in a <a>DAPT Document</a> by the following structure and constraints:</p>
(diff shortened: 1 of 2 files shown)


himorin:
I could not identify from where build error coming... It says html validation failed as:

>     "file:/home/runner/work/dapt/dapt.w3c/index.html":803.19618-803.19917: error: Attribute “title” not allowed on element “a” at this point.
>    "file:/home/runner/work/dapt/dapt.w3c/index.html":803.20843-803.21141: error: Attribute “title” not allowed on element “a” at this point.
>    "file:/home/runner/work/dapt/dapt.w3c/index.html":803.22877-803.23177: error: Attribute “title” not allowed on element “a” at this point.
 
but with exported html from respec, [nu html checker reported no error](https://validator.w3.org/nu/?doc=https%3A%2F%2Fttml.w3c.himor.in%2FED-dapt-20230822.html), nor I could not find title (not `xlink:title`) around reported line...
It seems similar to one we had initially with SVG image, but it does not change any SVG file at this point in this PR.

cconcolato:
@himorin I think the error is due to the SVG file referring to some sections of the spec that have been removed. Once the SVG is regenerated (my request to Nigel, above), that should be fine.

nigelmegitt:
Will do!

nigelmegitt:
Done, and the checks passed this time, confirming @cconcolato's hypothesis.

index.html
```diff
@@ -1675,9 +1616,13 @@ <h4>Time expressions</h4>
       </section> <!-- time-constraints-section -->
 
       <section>
-        <h4>Layout</h4>
+        <h4>Layout and styles</h4>
         <p>This specification does not put additional constraints on the layout and rendering <a>features</a> defined in [[!TTML-IMSC1.2]].</p>
         <div class="note">Layout of the paragraphs may rely on the default TTML region (i.e. if no <code>&lt;layout&gt;</code> element is used in the <code>&lt;head&gt;</code> element) or may be explicit by the use of the <code>region</code> attribute, to refer to a <code>&lt;region&gt;</code> element present at <code>/tt/head/layout/region</code>.</div>
+
+        <p>Style references or inline styles MAY be used, using <code>style</code> attributes 
+          and/or <code>&lt;style&gt;</code> elements as defined in [TTML2] or [TTML-IMSC1.2].</p>
```
nigelmegitt, line 1624: Need to describe how to do inline styles too, e.g.:

```suggestion
        <p>Style references or inline styles MAY be used, using any combination of
        <code>style</code> attributes, 
        <code>&lt;style&gt;</code> elements and
        inline style attributes as defined in [TTML2] or [TTML-IMSC1.2].</p>
```

nigelmegitt (approved):
Looks great, just one suggestion for completeness, but modulo that, approving.

andreastai (approved):
Looks good to me

PR #196, opened 2023-11-03
Add inline Registry Section

### LINKED ISSUE #195: Create inline Registry Tables and Registry Definition
We currently have two registries that are mentioned but do not exist. We have two choices to create them:
1. Separate document(s)
2. Embedded within DAPT alongside the Registry Definition section.

The current Process allows us to update Registry Tables embedded in Recommendations without going through the entire Rec track process again (or even a small part of it!). Suggest as initial attempt inlining the Registry tables into a single DAPT document.

nigelmegitt:
Closes #195. Incorporates and modifies the [TTWG boilerplate](https://github.com/w3c/ttwg/blob/fb28a74f733fd7803c7c7ad2eecc27852e4753b8/boilerplate/registry/index.html) registry definition, and specifies starter tables for `daptm:descType` and `daptm:eventType`.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/196.html" title="Last updated on Nov 23, 2023, 9:24 AM UTC (ef4ac7e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/196/9de2b7f...ef4ac7e.html" title="Last updated on Nov 23, 2023, 9:24 AM UTC (ef4ac7e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -7,9 +7,8 @@
     <script class="remove">
       // Rather than include the version control system (VCS) URL, use the Respec config value
       function replaceVCS(config) {
-        // If the element is an <a> already, Respec gives an error. Recreate the element as an <a> instead.
         document.querySelectorAll(".pp-vcs").forEach((element) => 
-          element.outerHTML=`<a class="${element.className}" href="${config.github.repoURL}">${element.innerHTML}</a>`);
+          element.href=config.github.repoURL)
       }
 
       var respecConfig = {
@@ -2701,7 +2700,7 @@ daptm:eventType : string
           <p>Changes to this  <a>W3C Registry</a> MUST be requested (the <dfn>change request</dfn>)
             using any one of the following options:</p>
           <ul>
-            <li>Open an issue on this document's <span class="pp-vcs">version control system</span>;</li>
+            <li>Open an issue on this document's <a class="pp-vcs" href="">version control system</a>;</li>
             <li>Send an email to the <a>TTWG</a>'s public email reflector
               <a href="mailto:public-tt@w3.org">public-tt@w3.org</a></li>
           </ul>
@@ -2717,7 +2716,7 @@ daptm:eventType : string
             <li>Any other supporting information, for example real world usage statistics;</li>
           </ul>
           <p>The proposer of the change MAY open a pull request (or equivalent) on
-            the <span class="pp-vcs">version control system</span>, with that pull request
+            the <a class="pp-vcs" href="">version control system</a>, with that pull request
             containing the proposed changes. If a pull request is opened then
             a corresponding issue MUST also be opened and
             the pull request MUST be linked to that issue.</p>
@@ -2732,7 +2731,7 @@ daptm:eventType : string
             <p>If the <a>custodian</a> is the <a>TTWG</a>:</p>
             <ul>
               <li>If the change proposer did not open a pull request
-                on the <span class="pp-vcs">version control system</span>,
+                on the <a class="pp-vcs" href="">version control system</a>,
                 then assessment is paused until a TTWG member
                 has opened such a pull request,
                 which MUST represent the requested changes
@@ -2756,7 +2755,7 @@ daptm:eventType : string
               </li>
             </ul>
             <p>An approved <a>change request</a> is enacted by merging its
-              related pull request into the <span class="pp-vcs">version control system</span>
+              related pull request into the <a class="pp-vcs" href="">version control system</a>
               and publishing the updated version of this document.</p>
           </section>
           <section id="team-change-request-assessment">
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Still to do: a few styling tweaks, and a note about registry sections in the document conventions.

nigelmegitt:
There will be some more tweaks to this based on https://lists.w3.org/Archives/Public/spec-prod/2023OctDec/0009.html and a W3C Slack response from @dbaron:

* rename Registry Section to Registry Definitions
* markup the Registry Definitions, which are not easily changeable without a review cycle, differently to the Registry Tables, whose entries are easily changeable.

There were other points, but I think they're less clear cut, and worth discussing more.

nigelmegitt:
Those tweaks made in 20380b0.

cconcolato (approved):
LGTM
Do we need section G.1.2.2.2 given that TTWG is the custodian of the defined registries?

nigelmegitt:
I plan to merge this on Nov 23 if there are no unresolved objections, since that will have been 2 weeks since this PR was properly ready for review.

index.html
```diff
@@ -2663,8 +2807,192 @@ <h3>#xmlLang-root</h3>
...
+            <li>Send an email to the <a>TTWG</a>'s public email reflector
+              <a href="mailto:public-tt@w3.org">public-tt@w3.org</a></li>
+          </ul>
+          <p>The <a>change request</a> MUST include enough information for the <a>custodian</a>
+            to be able to identify all of:</p>
+          <ul>
+            <li>The <a>registry table</a>;</li>
+            <li>The <a>registry entries</a> being proposed for addition;</li>
+            <li>The <a>provisional</a> <a>registry entries</a> for which changes are being proposed;</li>
+            <li>The <a>registry entries</a> being proposed for a <a>status</a> change;</li>
```
himorin, line 2843: quite minor, but above three items might be better to be quoted as `or`?

index.html
```diff
@@ -2663,8 +2807,192 @@ <h3>#xmlLang-root</h3>
...
+        Each <a>registry table</a> consists of a set of
+        <dfn data-cite="w3c-process#registry-entry" data-lt="registry entry|registry entries">registry entries</dfn>.</p>
+      <section id="registry-entries">
+        <h4>Registry Entries</h4>
+        <p>Each <a>registry entry</a> has a <a>status</a>, a unique <dfn>key</dfn>, and
+          if appropriate, other <dfn>fields</dfn>, for example any notes,
+          a description, or a reference to some other
+          defining entity.
+        </p>
+        <p>The <a>registry table</a> definition MUST define
```
himorin, line 2918: For here, isn't it better to link as `registry table definition` but not `registry table`? since MUST targets definition but not table itself...

himorin (approved):

himorin:
Added some minor questions, but overall seems fine for me.
One note, it seems working for now, but we may need to turn postProcess function into Promise if timing issue (not replaced within streamline publication) occurred.

index.html
nigelmegitt: I don't think so. I see this as three distinct lists, any of which could be empty.

If the lists are:
1. entries proposed for addition: `[foo]`
2. provisional entries being changed: `[bar]`
3. registry entries being proposed for a status change: `[]`

then it would not be acceptable to include only 1. or 2. but somehow obfuscate or hide the other.

index.html
nigelmegitt: I think you're right, that would be better. Will look to see if it causes any reference problems, and fix if I can.

nigelmegitt:
> we may need to turn postProcess function into Promise if timing issue (not replaced within streamline publication) occurred

How could there be a timing issue? As far as I can tell, everything else has already run beforehand, except maybe fixup.js which doesn't interact with this. I'd be more worried that making it async would _cause_ a timing issue.

index.html
nigelmegitt: Done in 03ab6c7.

index.html
himorin: I agree if this can accept empty input. My concern was combination of `MUST` clause and `identify all of`.
> The <a>change request</a> MUST include enough information for the <a>custodian</a> to be able to identify all of:

(yeah, identify should allow identified as empty..)

himorin:
> > we may need to turn postProcess function into Promise if timing issue (not replaced within streamline publication) occurred
> 
> How could there be a timing issue? As far as I can tell, everything else has already run beforehand, except maybe fixup.js which doesn't interact with this. I'd be more worried that making it async would _cause_ a timing issue.

In respec processing, if promise returned for postProcess, respec will wait to be resolved before going to next task.
https://lists.w3.org/Archives/Public/spec-prod/2017JanMar/0006.html (around mid)

index.html
nigelmegitt: I could change "The ... entries" to "Any ... entries" or "All ... entries" if that would make it even clearer?

nigelmegitt: Clarified in ef4ac7e as "All ... entries ..., if any" for each of those three affected bullets.

himorin: +1 on this! thank you.

Issue #188, opened 2023-10-05
Add a note that Character is not usually needed in AD scripts


nigelmegitt:
- Section 4.2 – Character
  - I didn’t get at first, that the entire Character section only applies to dubbing (even though the first sentence generally states this).  This might benefit from a note at the top of the section stating that ‘Character attributes do not apply to Audio Description scripts).  Then we AD people can just skip that section 😉.

_Originally posted by @nigelmegitt in https://github.com/w3c/dapt/discussions/187#discussioncomment-7199773_

nigelmegitt:
Pull request #208 open for review.

PR #208, opened 2024-01-31
Note use for no character identifier

### LINKED ISSUE #200: Note about association of Event with Character is incomplete
The note says:
> While typically, a Script Event corresponds to one single Character, there are cases where multiple characters can be associated with a Script Event. This is when all Characters speak the same text at the same time.

but the specification allows for *Zero or more* Character Identifiers.

I think we should highlight that no character is indicated when the event corresponds to in-image text or to audio descriptions.

nigelmegitt:
Closes #200 and #188.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/208.html" title="Last updated on Feb 15, 2024, 3:25 PM UTC (984e45e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/208/8d16753...984e45e.html" title="Last updated on Feb 15, 2024, 3:25 PM UTC (984e45e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -840,7 +840,13 @@ table.coldividers td + td { border-left:1px solid gray; }
             </li>
             <li>An optional <a>Script Event Type</a> used to identify the type of event (e.g. dialogue, on screen text) in <a>Dubbing Scripts</a>.</li>
             <li>Zero or more <a>Character Identifiers</a> indicating the <a>Characters</a> involved in this <a>Script Event</a>.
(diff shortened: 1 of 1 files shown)


cconcolato (commented):
Regarding the sentence:
"In Dubbing Scripts, it is necessary to identify each character in the programme. This is done with a Character object which has the following properties:"

I wonder if this should be changed to
"A character in the programme can be described using a Character object which has the following properties:."
Avoid "in Dubbing Scripts" because it could also be used in "Audio Descriptions"?
Avoid "it is necessary" which is too close to a normative statement.

cconcolato (approved):

PR #240, opened 2024-09-11
Add media zero timecode and start of programme timecode metadata

### LINKED ISSUE #232: Required metadata field for earliest SMPTE time code to allow conversion between DAPT and ESEF
Descriptions in ESEF files are timed via SMPTE timecodes so to convert to and from DAPT a key time code must be included in the metadata from which the timecodes of all other descriptions can be extrapolated. An existing EBUTT metadata field - documentStartOfProgramme - would be an obvious candidate for this but this value isn't typically recorded in an ESEF file. 

@nigelmegitt  proposes a new DAPT metadata value - equivalent to the ESEF format's first_content_tc - that all other relative values can be synchronised against.

nigelmegitt:
Closes #232.

Adds:
1. A metadata object called Media Zero Timecode
2. A metadata object called Start of Programme Timecode
3. An extension feature `#mediaZeroTimecode`, disposition Permitted
4. An example document fragment demonstrating the feature in use
5. An example algorithm for how to use the Media Zero Timecode and Start of Programme Timecode in practice to establish if the document is likely to be correctly synchronised with a related media object whose timings are timecode-based.

Would appreciate review comments from @ewanrsmith - I don't seem to be able to request a review from him.

**Updated 2026-09-25** to remove the `<legacy>` object and add the Start of Programme Timecode object

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/240.html" title="Last updated on Sep 28, 2024, 12:15 AM UTC (c1401ee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/240/f3fb76f...c1401ee.html" title="Last updated on Sep 28, 2024, 12:15 AM UTC (c1401ee)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1123,7 +1123,9 @@
               </ul>
               <p>When the related media's timecode is later dicovered,
                 the times of the <a>Script Events</a> can be adjusted so that
-                the zero points are coincident.
+                media time zero is coincident with the beginning of the
+                related media, and the <a>Media Zero Timecode</a> can be
+                removed or set to the now-known value.
               </p>
             </div>
           </section>
--- a/index.html
+++ b/index.html
@@ -2120,7 +2120,7 @@ daptm:descType : string
       </section>
 
       <section>
-        <h3>Related Media Object (TTML)</h3>
+        <h3>Related Media Object</h3>
         <p>Within DAPT, the common language terms audio and video are used in the context of a programme.
           The audio and video are each a part of what is defined in [[TTML2]] as the
           <dfn data-cite="TTML2#terms-related-media-object">Related Media Object</dfn> that
@@ -2774,8 +2774,6 @@ daptm:descType : string
       during presentation.
     </p>
 
-    <p>A <a>DAPT Script</a> MAY contain zero or one <a>DAPT Origin Timecode</a> objects.</p>
-    <p>A <a>DAPT Script</a> MAY contain zero or one <a>Start of Programme Timecode</a> objects.</p>
     <p>If either a <a>DAPT Origin Timecode</a> object
       or a <a>Start of Programme Timecode</a> object is present,
       the <a>DAPT Document</a> MUST have
--- a/index.html
+++ b/index.html
@@ -2764,7 +2764,7 @@ daptm:descType : string
       a common issue occurs if the timecode of
       the beginning of the media is not known at the time of conversion.
     </p>
-    <p>The optional <a>Media Zero Timecode</a> and
+    <p>The optional <a>DAPT Origin Timecode</a> and
       <a>Start of Programme Timecode</a> properties can be used to
       identify when there is a possible synchronisation error,
       and to resynchronise the document when all
@@ -2774,9 +2774,9 @@ daptm:descType : string
       during presentation.
     </p>
 
-    <p>A <a>DAPT Script</a> MAY contain zero or one <a>Media Zero Timecode</a> objects.</p>
+    <p>A <a>DAPT Script</a> MAY contain zero or one <a>DAPT Origin Timecode</a> objects.</p>
     <p>A <a>DAPT Script</a> MAY contain zero or one <a>Start of Programme Timecode</a> objects.</p>
-    <p>If either a <a>Media Zero Timecode</a> object
+    <p>If either a <a>DAPT Origin Timecode</a> object
       or a <a>Start of Programme Timecode</a> object is present,
       the <a>DAPT Document</a> MUST have
       one <code>&lt;head&gt;</code> element child of the <code>&lt;tt&gt;</code> element,
@@ -2784,29 +2784,29 @@ daptm:descType : string
       at least one <code>&lt;metadata&gt;</code> element child.</p>
 
     <section>
-      <h5>Media Zero Timecode</h5>
-      <p>The optional <dfn>Media Zero Timecode</dfn> allows a timecode value to be declared
+      <h5>DAPT Origin Timecode</h5>
+      <p>The optional <dfn>DAPT Origin Timecode</dfn> allows a timecode value to be declared
         that corresponds to the zero point on the media timeline,
         that is, the time of a hypothetical <a>Script Event</a> whose <a>Begin</a> is
         zero seconds on the media timeline.</p>
-      <p>A <a>DAPT Script</a> MAY contain zero or one <a>Media Zero Timecode</a> objects.</p>
-      <p>The <a>Media Zero Timecode</a> object is represented in a <a>DAPT Document</a> by
-        a <code>&lt;</code><dfn><code>daptm:mediaZeroTimecode</code></dfn><code>&gt;</code> element present at the path
-        <code>/tt/head/metadata/daptm:mediaZeroTimecode</code>,
+      <p>A <a>DAPT Script</a> MAY contain zero or one <a>DAPT Origin Timecode</a> objects.</p>
+      <p>The <a>DAPT Origin Timecode</a> object is represented in a <a>DAPT Document</a> by
+        a <code>&lt;</code><dfn><code>daptm:daptOriginTimecode</code></dfn><code>&gt;</code> element present at the path
+        <code>/tt/head/metadata/daptm:daptOriginTimecode</code>,
         with the following constraints:</p>
       <ul>
         <li>The character content of the element must conform to a <code>clock-time</code>
           with a <code>frames</code> component, as defined in <dfn data-cite="ttml2#timing-value-time-expression">&lt;time-expression&gt;</dfn>.</li>
       </ul>
       <p class="note">See also <a href="#ttp-framerate"></a>.
-        No mechanism is defined here for declaring a different frame rate for the <a>Media Zero Timecode</a>
+        No mechanism is defined here for declaring a different frame rate for the <a>DAPT Origin Timecode</a>
         than is used for other frame-based time expressions.
       </p>
       <pre class="example">
 ...
 &lt;head&gt;
 &lt;metadata&gt;
-&lt;daptm:mediaZeroTimecode&gt;10:01:20:12&lt;/daptm:mediaZeroTimecode&gt;
+&lt;daptm:daptOriginTimecode&gt;10:01:20:12&lt;/daptm:daptOriginTimecode&gt;
 &lt;/metadata&gt;
 &lt;/head&gt;
 &lt;body&gt;
@@ -2820,11 +2820,11 @@ daptm:descType : string
         <p>When converting legacy formats that store the equivalent of each Script Event,
           for example a <a>description</a>, with SMPTE timecodes,
           but where the <a>Start of Programme Timecode</a> is
-          not stored, and is not immediately available, the <a>Media Zero Timecode</a>
+          not stored, and is not immediately available, the <a>DAPT Origin Timecode</a>
           can be used to defer synchronisation with the media until
           that media begin timecode is known.
         </p>
-        <p>One approach to deferring synchronisation using <a>Media Zero Timecode</a>
+        <p>One approach to deferring synchronisation using <a>DAPT Origin Timecode</a>
           is as follows, assuming that a legacy source file is being converted
           to a <a>DAPT Document</a> and relates to media with continuous timecode:
         </p>
@@ -2832,12 +2832,12 @@ daptm:descType : string
           <li>Find the earliest begin timecode for an item in the source file
             that will be mapped to a <a>Script Event</a>;</li>
           <li>Create a <a>DAPT Document</a> with
-            a <a>Media Zero Timecode</a> set to that earliest begin timecode;</li>
+            a <a>DAPT Origin Timecode</a> set to that earliest begin timecode;</li>
           <li>For each relevant item in the source file,
             create a <a>Script Event</a> with appropriate
             <a>Begin</a> and <a>End</a> times obtained by converting
             the source file's begin and end timecodes to media time,
-            relative to that <a>Media Zero Timecode</a>,
+            relative to that <a>DAPT Origin Timecode</a>,
             basing elapsed time on the applicable frame rate;</li>
           <li>If using frame-based media time expressions,
             set the <code>ttp:frameRate</code> attribute of
@@ -2847,7 +2847,7 @@ daptm:descType : string
         <p>When the related media's timecode is later dicovered,
           the times of the <a>Script Events</a> can be adjusted so that
           media time zero is coincident with the <a>Start of Programme Timecode</a>,
-          and the <a>Media Zero Timecode</a> can be
+          and the <a>DAPT Origin Timecode</a> can be
           removed or set to the now-known value.
         </p>
         <p class="note">The above algorithm can be used to align the begin times
@@ -2862,11 +2862,11 @@ daptm:descType : string
       <h5>Start of Programme Timecode</h5>
       <p>The optional <dfn>Start of Programme Timecode</dfn> allows a timecode value to be declared
         that corresponds to the beginning of the related media object's programme content.</p>
-      <p >In combination with <a>Media Zero Timecode</a>,
+      <p >In combination with <a>DAPT Origin Timecode</a>,
         the value of <a>Start of Programme Timecode</a> can be used to infer whether or not
         the media times in the <a>DAPT Script</a> are likely to be correctly synchronised with
         the <a>Related Media Object</a>.</p>
-      <p>If both <a>Media Zero Timecode</a> and <a>Start of Programme Timecode</a> are present,
+      <p>If both <a>DAPT Origin Timecode</a> and <a>Start of Programme Timecode</a> are present,
         but their values are different, it is likely that the media times are not
         synchronised with the <a>Related Media Object</a>, since this implies that
         the equivalent time code to zero seconds in media time is not the start of the programme,
@@ -2880,7 +2880,7 @@ daptm:descType : string
 ...
 &lt;head&gt;
 &lt;metadata&gt;
-&lt;daptm:mediaZeroTimecode&gt;10:01:20:12&lt;/daptm:mediaZeroTimecode&gt;
+&lt;daptm:daptOriginTimecode&gt;10:01:20:12&lt;/daptm:daptOriginTimecode&gt;
 &lt;ebuttm:documentStartOfProgramme&gt;10:00:00:00&lt;ebuttm:documentStartOfProgramme&gt;
 &lt;!-- It is likely that this document is 1 minute, 20 seconds and 12 frames too early,
   compared to the related media --&gt;
@@ -2895,7 +2895,7 @@ daptm:descType : string
       </pre>
       <p>If the times of the <a>Script Events</a> are adjusted to bring
         the media time into synchronisation with the <a>Related Media Object</a>,
-        as noted in <a>Media Zero Timecode</a>, the <a>Start of Programme Timecode</a>
+        as noted in <a>DAPT Origin Timecode</a>, the <a>Start of Programme Timecode</a>
         SHOULD NOT be changed, since it is an invariant feature of the <a>Related Media Object</a>,
         and does not describe the times in the <a>DAPT Document</a>.</p>
     </section>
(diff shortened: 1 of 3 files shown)


mattsimpson3:
Makes sense to me.

mattsimpson3 (commented):
Nothing to change from me...

mattsimpson3 (approved):
Approved this time!

cconcolato:
Why do we need a `legacy` element? Why don't we use an attribute like all the EBU-TT equivalent metadata?

cconcolato:
> The idea is to clearly mark out data that is not core to the content, and is definitely not expected to be required or processed for presentation, 

That is already conveyed with the metadata signaling (qualified namespace on the attribute or when used on the element).

> but is only needed to retain information that was part of the legacy source file, which may need to be resolved later. I'd happily add some descriptive text to try to explain this more. We could consider other names too.

This raises a lot of questions for me:
- what is "legacy"? how do you define it? For example, would Netflix TTAL count as legacy format? Would you expect TTAL converted documents to be in the legacy element? The name does not matter much, it's the semantics that are required.
- We have an example for adding vendor metadata in a `metadata` element, example 32. What's different here? We would need clear semantics as to what goes into `legacy` and what does not. I think it's calling for unnecessary complexity.
- In which `metadata` element should this legacy go? I wouldn't want to have it in the same one as the `ttm:agent` elements that represent characters. 

> Re EBU-TT metadata, which is in [EBU-TT part M, Metadata Definitions, EBU Tech 3390](https://tech.ebu.ch/publications/tech3390), defines most of the entities as elements, not attributes: 

> doing this means that if anyone needs to add some custom metadata they have the ability to add attributes to the element.

Understood, but `ebuttm:documentStartOfProgramme` which is similar is an attribute, not an element.

> doing this means that if anyone needs to add some custom metadata they have the ability to add attributes to the element.

It could make sense for some structured vocabulary, but in this case where there is just one timecode value to carry, defining 2 elements is too complex.

> An alternative that you mentioned in yesterday's call is to define an attribute that would live on the root <tt> element. I think that would increase the likelihood that folk would get used to it being there, and consider it as somehow important for processing the document in all cases.  It would certainly be possible to do it this way, but I have a strong preference for avoiding it.

I don't understand why you think people would think it would needed in all cases, if it is optional. It feels natural to me to have document-wide attributes be on the top-level element. 

My strong preference is to have it as an attribute, and not to introduce unnecessary elements. `ebuttm:documentStartOfProgramme` is on `/tt/head/metadata`, why not use it there? Alternatively, since it relates to the timeline of the document, we could have it on the `body` element.

cconcolato:
Some additional comments:
> the time of a [Script Event](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/240.html#dfn-script-event) whose [Begin](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/240.html#dfn-begin) is zero seconds on the media timeline.

This gives the impression that there is always a Script Event at time 0. It should be rephrased to say things like 'hypothetical'.

> One approach to resolving this conundrum 

The word 'conundrum' seems too strong. It is a rather straightforward algorithm.

nigelmegitt:
> Some additional comments:
> 
> > the time of a [Script Event](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/240.html#dfn-script-event) whose [Begin](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/240.html#dfn-begin) is zero seconds on the media timeline.
> 
> This gives the impression that there is always a Script Event at time 0. It should be rephrased to say things like 'hypothetical'.

Good catch, thank you.

> 
> > One approach to resolving this conundrum
> 
> The word 'conundrum' seems too strong. It is a rather straightforward algorithm.

It's a conundrum for the person converting the legacy (e.g. ESEF) files, who has no access to the data they apparently need.

nigelmegitt:
> Understood, but `ebuttm:documentStartOfProgramme` which is similar is an attribute, not an element.

That's not correct, it's an element.

> > doing this means that if anyone needs to add some custom metadata they have the ability to add attributes to the element.
> 
> It could make sense for some structured vocabulary, but in this case where there is just one timecode value to carry, defining 2 elements is too complex.

Right, one _so far_...

> > An alternative that you mentioned in yesterday's call is to define an attribute that would live on the root `<tt>` element. I think that would increase the likelihood that folk would get used to it being there, and consider it as somehow important for processing the document in all cases.  It would certainly be possible to do it this way, but I have a strong preference for avoiding it.
> 
> I don't understand why you think people would think it would be needed in all cases, if it is optional. It feels natural to me to have document-wide attributes be on the top-level element.

Unfortunately I've seen plenty of misunderstandings and weird/unhelpful/not-as-intended documents from people who haven't taken the time to read the spec properly. I think the fear is reasonably well founded.

> My strong preference is to have it as an attribute and not to introduce unnecessary elements. `ebuttm:documentStartOfProgramme` is on `/tt/head/metadata`, why not use it there? Alternatively, since it relates to the timeline of the document, we could have it on the `body` element.

Yes, we could put it there instead.

> > The idea is to clearly mark out data that is not core to the content, and is definitely not expected to be required or processed for presentation,
> 
> That is already conveyed with the metadata signaling (qualified namespace on the attribute or when used on the element).

Yes, to an extent, but again, my experience is that people look past namespaces.

> > but is only needed to retain information that was part of the legacy source file, which may need to be resolved later. I'd happily add some descriptive text to try to explain this more. We could consider other names too.
> 
> This raises a lot of questions for me:
> 
>     * what is "legacy"? how do you define it? For example, would Netflix TTAL count as legacy format? Would you expect TTAL converted documents to be in the legacy element? The name does not matter much, it's the semantics that are required.

Since its metadata, this isn't so much about semantics as about guidance and real world usage. In my way of thinking, data should be put here if it is a) not otherwise needed/used/represented in DAPT, and b) still needed for some later purpose because of the workflow, or needed to support the transition to DAPT.

>     * We have an example for adding vendor metadata in a `metadata` element, example 32. What's different here? We would need clear semantics as to what goes into `legacy` and what does not. I think it's calling for unnecessary complexity.

Yes, we could omit this altogether and allow people to define and use their own custom data; the reason for trying to resolve this issue within the DAPT specification is because we expect it to be a widespread issue for people coming from ESEF in particular.

>     * In which `metadata` element should this legacy go? I wouldn't want to have it in the same one as the `ttm:agent` elements that represent characters.

OK, so you _are_ in favour of organising the metadata in some way. Doesn't the addition of a legacy metadata child of `<metadata>` help with that? Why would you not want to use the same `<metadata>` element for multiple purposes? Most of the time I don't think TTML implementers even realise that it's possible to have more than one `<metadata>` child of the same parent element.

ewanrsmith:
I think adding a legacy metadata element seems like a cleaner approach, given that it will function as a container for values that will only be meaningful in the context it represents:

- mediaZeroTimecode would allow conversion from exchange formats timed using SMPTE timecodes into DAPT such that the synchronisation of the events is preserved relative to one another, but not to the start of the media. This assumes the absence of the ebutt:documentStartOfProgramme value (or equivalent) at the time of conversion, as otherwise the media time would have been set against it.

- ebutt:documentStartOfProgramme would be added to a document that was converted based on mediaZeroTimecode and would provide an offset value based on the difference between the two fields. Its presence would also differentiate between documents for which the synchronisation has been validated and those for which it may be unclear.

Ordinarily my instinct would be to add them as attributes for the reasons Cyril suggests, but I think it makes sense to group them under an element that can dispensed with for natively originated documents. I think anyone dealing with legacy files in volume at the time of adoption would appreciate having these two values logically grouped together but separately.

cconcolato:
> > Understood, but `ebuttm:documentStartOfProgramme` which is similar is an attribute, not an element.
> 
> That's not correct, it's an element.

I stand corrected. I guess it feels so unnatural to me to use elements for that that I persuaded myself it was an attribute ...


> 
> > > doing this means that if anyone needs to add some custom metadata they have the ability to add attributes to the element.
> > 
> > 
> > It could make sense for some structured vocabulary, but in this case where there is just one timecode value to carry, defining 2 elements is too complex.
> 
> Right, one _so far_...

I meant: one for `legacy` and one for `mediaZeroTimecode` vs. 0 for a 'mediaZeroTimecode' attribute.

> > This raises a lot of questions for me:
> > ```
> > * what is "legacy"? how do you define it? For example, would Netflix TTAL count as legacy format? Would you expect TTAL converted documents to be in the legacy element? The name does not matter much, it's the semantics that are required.
> > ```
> 
> Since its metadata, this isn't so much about semantics as about guidance and real world usage. 

It does not feel sufficient to introduce an element and just provide guidance on how to use it without clear normative semantics.

> In my way of thinking, data should be put here if it is a) not otherwise needed/used/represented in DAPT, and b) still needed for some later purpose because of the workflow, or needed to support the transition to DAPT.

I'd rather have a stricter set of rules, like do not put additional vocabulary anywhere except in the legacy element. Otherwise,  an implementation will have to deal with both approaches ...

> 
> > ```
> > * We have an example for adding vendor metadata in a `metadata` element, example 32. What's different here? We would need clear semantics as to what goes into `legacy` and what does not. I think it's calling for unnecessary complexity.
> > ```
> 
> Yes, we could omit this altogether and allow people to define and use their own custom data; the reason for trying to resolve this issue within the DAPT specification is because we expect it to be a widespread issue for people coming from ESEF in particular.

The use case of ESEF can be solved in many ways. If we provide a clear example, people will follow it. We could also go with strict semantics.

> 
> > ```
> > * In which `metadata` element should this legacy go? I wouldn't want to have it in the same one as the `ttm:agent` elements that represent characters.
> > ```
> 
> OK, so you _are_ in favour of organising the metadata in some way. 

Yes, but I'd like implementations to be simple.

> Doesn't the addition of a legacy metadata child of `<metadata>` help with that? 

It would if a) it had clear, normative semantics to make it unambiguous for authors and simple to implement for parsers. and b) probably the name was changed, like `external` (meaning anything non-DAPT). But I'd still prefer tagging the metadata element to indicate the type of metadata.

> Why would you not want to use the same `<metadata>` element for multiple purposes? 

It seems easier to discard an entire metadata element than to check the type of every child element.

> Most of the time I don't think TTML implementers even realise that it's possible to have more than one `<metadata>` child of the same parent element.

We already have a sentence that says: 
> For example, to refer to the first `<metadata>` element child of the `<head>` element child of the `<tt>` element, the following path would be used: /tt/head/metadata[0].

cconcolato:
Should mandate that there be at most 1 metadata in head and in script events?

nigelmegitt:
I'm not sure I've ever seen a use case for multiple `<metadata>` element children of the same parent, I think it would be fine to restrict this.

nigelmegitt:
> Should we mandate that there be at most 1 metadata in head and in script events?

Raised as #242.

index.html
```diff
@@ -886,6 +886,152 @@ <h5>Timing Properties</h5>
...
+          </p>
+          <p>The optional <a>Media Zero Timecode</a> and
+            <a>Start of Programme Timecode</a> properties can be used to
+            identify when there is a possible synchronisation error,
+            and to resynchronise the document when all
+            the required information is known.</p>
+          <p>These properties are provided as metadata only and
+            are not intended to be used to perform direct synchronisation offsets
+            during presentation.
+          </p>
```
cconcolato, line 906: I'm still trying to understand the feature here and how it differs from the existing EBU-TT feature.

Is the proposed text trying to say the following:
> As defined in TTML2, each DAPT document has a timeline on which Script Events are positioned using their begin attributes. For synchronization purposes it is useful to map the DAPT document timeline with the related media object timeline.
In some cases, the mapping is simple, e.g. media time 0 in the related media object corresponds to the origin of the DAPT document timeline. In other cases, a timecode may be necessary. DAPT provides the ability to provide 2 different timecodes: 
> * The "Start of the Programme Timecode" is the timecode of the time instant in the related media object corresponding to the origin of the DAPT document timeline. It is represented in a DAPT document by a documentStartOfProgramme element, defined in EBUT-TT, using the EBU-TT namespace, located in a metadata element in the head. 
>
> (and either one of the alternatives below)
> * The "Media Zero Timecode" is the timecode of the first media data in the related media object. It is represented in DAPT documents by the mediaZeroTimecode element, in the DAPT namespace, located in a metadata element in the head.
> * The "First Script Event Timecode" is the timecode of the time instant in the related media object corresponding to the first script event in the DAPT document. It is represented in DAPT documents by the XXX element, in the DAPT namespace, located in a metadata element in the head.
>
> NOTE: Depending on the script and related media object workflows, there may not be a timecode in the related media object corresponding to the origin of the DAPT document timeline, for example if the related media object has been edited to remove some data at its start. In that case, either the "Start of the Programme Timecode" and all the Script Event begin times have to be edited, or alternatively the "Start of the Programme Timecode" cannot be used but the "First Script Event Timecode" or "Media Zero Timecode" can be used.

index.html
```diff
@@ -886,6 +886,152 @@ <h5>Timing Properties</h5>
             </p>
           </section>
         </section>
+
+        <section>
+          <h4>Deferred resolution of timecode-based synchronisation</h4>
```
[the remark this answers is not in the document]
nigelmegitt: I think it makes sense to move it to an appendix.

index.html
nigelmegitt: No, very much not that!

Start of programme timecode is the timecode in the _related media_ corresponding to the beginning of the related media object's programme content.

Media Zero timecode is a timecode that media time zero in the _DAPT document_ maps to.

If the DAPT document is not synchronised to the related media object's programme content, those two values will differ.

If the start of programme timecode is unknown, then you don't know if they're in sync or not.

If both are present and the same, then you have some reassurance that the DAPT document is synchronised with the media.

nigelmegitt: >the timecode of the time instant in the related media object corresponding to the origin of the DAPT document timeline

The issue is that documents that are not synchronised to the related media object do not satisfy the "corresponding to the origin of the DAPT document timeline" part of this sentence.

nigelmegitt:
In discussion with @cconcolato we realised that Media Zero Timecode was confusing, since it wasn't obvious _which_ media was meant - it needs to mean "media timebase in the DAPT document" not "timecode at the beginning of the media". Changed to "DAPT Origin Timecode" in ed1f5de.

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
 
   </section>
 
+  <section class="appendix">
+    <h4>Timecode-related metadata</h4>
+    <p><a>DAPT Documents</a> express time as <em>media time</em>,
+      which assumes that there is a reference start time (zero) on a media
+      timeline, and a fixed playrate.
+      An alternative scheme that is used in some workflows is to
```
cconcolato, line 2627: I think the wording is ambiguous. Some workflows that use DAPT? No, I think it is meant to say: "used for some other script formats than DAPT"

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+      synchronise media components using timecode,
+      time stamps that are applied, for example to each frame of video.
```
cconcolato, line 2629: Is "time stamps that are applied" intended? It feels like left over text. Suggest delete it.
```suggestion
      for example to each frame of video.
```

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+    </p>
+    <p>Workflows that create <a>DAPT documents</a> from
+      resources that use timecode need to map the timecode
```
cconcolato, line 2632: "resources" can be confused with media resources (MPEG-2 TS, ProRes, ...) that can still use timecodes.

```suggestion
      script formats other than DAPT that use timecode need to map the timecode
```

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+      values onto the <a>DAPT document</a>'s timeline.</p>
+    <p>If this mapping is not correct, presentation of the
+      <a>DAPT Document</a> will not be synchronised with the <a>related media object</a>.
+      In order to achieve correct synchronisation in this scenario,
+      the timecode corresponding to the start of the programme needs to be known.</p>
```
cconcolato, line 2637: > the timecode corresponding to the start of the programme needs to be known.

This is too strong. Having one timecode is sufficient. It could be the first content timecode.  

```suggestion
       a timecode, such as the one corresponding to the start of the programme, needs to be known.</p>
```

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+    <p>If the start of programme timecode is not known,
+      but timecodes corresponding to <a>Script Events</a> are known,
+      it is still possible to construct a <a>DAPT Document</a>,
+      albeit one whose synchronisation with the related media is uncertain.</p>
```
cconcolato, line 2641: What do you expect the person reading "uncertain" to understand?

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+    <p>The optional <a>DAPT Origin Timecode</a> and
+      <a>Start of Programme Timecode</a> properties can be used to
+      identify when there is a possible synchronisation error,
+      and to resynchronise the document when all
+      the required information is known.</p>
+    <p>These properties are provided as metadata only and
+      are not intended to be used to perform direct synchronisation offsets
+      during presentation.</p>
```
cconcolato, line 2649: We should add:
In particular, when the related media object uses timecode, the presence of the timecode properties does not mean that the player needs to relate these timecode values with any timecode value embedded in the related media resource.

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+      the timecode corresponding to the start of the programme needs to be known.</p>
+    <p>If the start of programme timecode is not known,
+      but timecodes corresponding to <a>Script Events</a> are known,
+      it is still possible to construct a <a>DAPT Document</a>,
+      albeit one whose synchronisation with the related media is uncertain.</p>
+    <p>The optional <a>DAPT Origin Timecode</a> and
+      <a>Start of Programme Timecode</a> properties can be used to
+      identify when there is a possible synchronisation error,
+      and to resynchronise the document when all
+      the required information is known.</p>
```
cconcolato, line 2646: If my understanding is correct I would suggest replacing with the following text

> The properties can be used to provide timecodes from the related media object or from any other script format used to produce the DAPT document, and are informational. However, when they are both provided and differ, it is an indication that the DAPT document is not synchronized with the related media object and that processing of the script event begins is needed. To achieve synchronization, the following needs to be done:
> - The difference DAPT Origin Timecode minus the Start of Programme Timecode is computed as "delta". It may be positive or negative.
> - Each Script begin value X shall be changed to X + delta.
> - The DAPT Origin Timecode shall be removed or changed to the Start of Programme timecode value.

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+      or a <a>Start of Programme Timecode</a> object is present,
+      the <a>DAPT Document</a> MUST have
+      one <code>&lt;head&gt;</code> element child of the <code>&lt;tt&gt;</code> element,
+      and that <code>&lt;head&gt;</code> element MUST have
+      at least one <code>&lt;metadata&gt;</code> element child.</p>
+
+    <section>
+      <h5>DAPT Origin Timecode</h5>
+      <p>The optional <dfn>DAPT Origin Timecode</dfn> allows a timecode value to be declared
+        that corresponds to the zero point on the media timeline,
```
cconcolato, line 2661:
```suggestion
        that corresponds to the zero point of the DAPT document timeline,
```

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+        that is, the time of a hypothetical <a>Script Event</a> whose <a>Begin</a> is
+        zero seconds on the media timeline.</p>
```
cconcolato, line 2663:
```suggestion
        zero seconds.</p>
```

index.html
```diff
@@ -2610,6 +2619,161 @@ <h2>Security Considerations</h2>
...
+        <code>/tt/head/metadata/daptm:daptOriginTimecode</code>,
+        with the following constraints:</p>
+      <ul>
+        <li>The character content of the element must conform to a <code>clock-time</code>
+          with a <code>frames</code> component, as defined in <dfn data-cite="ttml2#timing-value-time-expression">&lt;time-expression&gt;</dfn>.</li>
+      </ul>
+      <p class="note">See also <a href="#ttp-framerate"></a>.
+        No mechanism is defined here for declaring a different frame rate for the <a>DAPT Origin Timecode</a>
+        than is used for other frame-based time expressions.
+      </p>
```
cconcolato, line 2676:
```suggestion
      </p>
      <p>If the related media object contains a timecode for the video frame synchronized to the origin of the DAPT document timeline, the DAPT origin timecode shall be equal that timecode. 
      </p>
```

cconcolato (approved):
Approving it technically, may need to do some editorial changes during CR phase.

PR #241, opened 2024-09-12
Remove Script Event Type, use Represents instead

### LINKED ISSUE #227: Interaction between "represents" and "script event type"
As discussed in the TTWG call on April 11, 2024, during the review of #217, we wonder what the interactions between the new `daptm:represents` and `daptm:eventType`. Both attributes are registry-based.

nigelmegitt:
Closes #227 by removing Script Event Type, reusing Represents instead.

* Merge registry table entries for `eventType` and `<content-descriptor>` (removing the `eventType` registry)
* Extract definition of Represents into the Shared Properties section
* Remove Script Event Type
* Make Represents an optional property of Script Event and move example that was in Script Event Type up into Script Event under the Represents bullet
* Add _permitted_ `#represents-div` feature
* Update class diagram to remove Script Event Type and add Represents to Script Event

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/241.html" title="Last updated on Sep 28, 2024, 12:12 AM UTC (5a3704d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/241/0d083c2...5a3704d.html" title="Last updated on Sep 28, 2024, 12:12 AM UTC (5a3704d)">Diff</a>


cconcolato:
If verbosity was not a concern, we should say:
1. a Script Event shall have a non-empty `daptm:represents` attribute and maybe, that contains a single value 
2. a Script Event shall not use a value that is not listed directly or indirectly (because it is a subset of a value listed) on the document-level `daptm:represents`. In the meeting @nigelmegitt said that there could be values listed at the top level that are not listed in script events, e.g. after editing or to indicate that the document tried to represent that dimension but could not find any content. 

We could live with that. Alternatively, we could also specify that either:
1. if `daptm:represents` is not present on a Script Event, it inherits the first value of the `daptm:represents` attribute on the document. It would behave like the default language.
2. or introduce a `daptm:defaultRepresents` that is used to avoid the use of "first value".

BTW, the property is mandatory on the DAPT script but it shall also be non-empty.

nigelmegitt:
@cconcolato see also https://github.com/w3c/dapt/issues/227#issuecomment-2348776405

I'm happy to take the approach you're suggesting - it's one of the options I also considered:

1. Make `daptm:represents` mandatory and non-empty and maybe single value on Script Events
2. Define a sub-type hierarchy in the `<content-descriptor>` registry. I would prefer to constrain it to 2 levels to minimise implementation complexity, i.e. to prevent the need to generate a tree, which would happen if we permitted a value to be a sub-type of another value that is also a sub-type of a 3rd one, and so forth.

If verbosity is a concern, perhaps we could shorten the attribute name, e.g to `repr` which is an abbreviation for "represents" or "representation" used in Python and Rust.

cconcolato:
> If verbosity is a concern, perhaps we could shorten the attribute name, e.g to repr which is an abbreviation for "represents" or "representation" used in Python and Rust.

The verbosity concern is not really about the length of the attribute name, but more about the repetition of the attribute in the case the document `represents` contains a single value that matches all the values on all the script events. But that concern is probably outweighed by the clear identification of what div matches a script event.

index.html
```diff
@@ -885,6 +831,103 @@ <h5>Timing Properties</h5>
               The end time of a <a>DAPT Script</a> is for practical purposes the end of the <a>Related Media Object</a>.
             </p>
           </section>
+
+          <section id="sharedprop-represents">
+            <h5>Represents</h5>
+            <p>The <dfn>Represents</dfn> property
+              indicates which components of the <a>related media object</a>
+              the contents of the parent object represent.
```
cconcolato, line 839: "parent object" means "current object" right?

index.html
```diff
@@ -885,6 +831,103 @@ <h5>Timing Properties</h5>
...
+              are listed in the following <a>registry table</a>:</p>
+              <div class="registry-table-section">
+                <table class="data">
+                  <caption><a>Registry table</a> for the <code>&lt;content-descriptor&gt;</code> component
+                    whose Registry Definition is at <a href="#registry-table-content-descriptor"></a></caption>
+                  <thead>
+                    <th><code>&lt;content-descriptor&gt;</code></th>
+                    <th>Status</th>
+                    <th>Description</th>
+                    <th>Used in</th>
```
cconcolato, line 874: What does "used in" mean? What are the normative implications, if any? Should we say "Example use case" to make it clear that has no normative meaning?

cconcolato (commented):
My general comments are:
- technical:  why is 'represents' optional on Script Events? If it remains optional, what should be its value when not specified? I would prefer making it mandatory.
- technical: I would like to mandate that if a represents value is used at a Script Event level, it (or its corresponding top-level value, if the value is a sublevel) must be present at the document level.
- editorial: it is not clear to me why we are making it a "shared property" like begin/end. Those are allowed on plenty of objects (Script Event, Mixing Instruction, Audio Recording). Represents is allowed on DAPT Script and Script Event. Language is allwoed on Script Event Description, Text (and also on Script Event because of inheritance). Text Language Source is allowed on DAPT Script, Script Event (by inheritance), Text. I think we should be consistent. My suggestion is maybe to remove the notion of "Shared properties", define a property in one place and allow it in other places by reference.

I added some other minor comments.

index.html
nigelmegitt: Changed to Example Usage.

index.html
nigelmegitt: This has changed now.

nigelmegitt:
This doesn't build properly because of https://github.com/speced/respec/issues/4806

nigelmegitt:
@cconcolato I think all of your comments have now been resolved, please check.

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                    <td><code>visual;text</code></td>
+                    <td>Yes</td>
+                  </tr>
+                  <tr>
+                    <td><code>visual;text</code></td>
+                    <td><code>visual;text;location</code></td>
+                    <td>yes</td>
+                  </tr>
+                </tbody>
+              </table>
```
cconcolato, line 918: This feels too complicated to me. Can we simply say:
Some descriptor values may be a prefix to other values, when these other values are more precise. For example, the value "visual" would be used when a script event represents anything the video part of the programme, while "visual;text" would be used only when the script event represents only text content appearing in the video part of the programme.

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                      <td><code>visual;text;location</code></td>
+                      <td>Provisional</td>
+                      <td>A <a>sub-type</a> of <code>visual;text</code> where the text indicates the location where the content is occurring.</td>
+                      <td></td>
+                      <td></td>
+                    </tr>
+                  </tbody>
+                </table>
+              </div>
+              <p>User-defined values MUST begin with <code>x-</code>.</p>
```
cconcolato, line 1009: So user-defined values for a visual part would have to be `x-visual;myValue`? Wouldn't it be better to allow `visual;x-myValue`?

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                    <th>Status</th>
+                    <th>Description</th>
+                    <th>Example usage</th>
+                    <th>Notes</th>
+                  </thead>
+                  <tbody>
+                    <tr>
+                      <td><code>audio</code></td>
+                      <td>Provisional</td>
+                      <td>Audio content</td>
```
cconcolato, line 939:
```suggestion
                      <td>Indicates that the DAPT content represents any part of the audio programme</td>
```

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                      <td>Dubbing, translation and hard of hearing subtitles and captions, pre- and post- production scripts</td>
+                      <td>See sub-types.</td>
+                    </tr>
+                    <tr>
+                      <td><code>audio;dialogue</code></td>
+                      <td>Provisional</td>
+                      <td>Verbal communication, in audio</td>
```
cconcolato, line 946:
```suggestion
                      <td>Indicates that the DAPT content represents verbal communication in the audio programme</td>
```

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                      <td>Dubbing, translation and hard of hearing subtitles and captions, pre- and post- production scripts</td>
+                      <td>For example, a spoken conversation.</td>
+                    </tr>
+                    <tr>
+                      <td><code>audio;nonDialogueSounds</code></td>
+                      <td>Provisional</td>
+                      <td>Sounds that are not verbal communication, in audio</td>
```
cconcolato, line 953:
```suggestion
                      <td>Indicates that the DAPT content represents a part of the audio programme corresponding to sounds that are not verbal communication</td>
```

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                      <td>Translation and hard of hearing subtitles and captions, pre- and post- production scripts</td>
+                      <td>For example, significant sounds, such as a door being slammed in anger.</td>
+                    </tr>
+                    <tr>
+                      <td><code>visual</code></td>
+                      <td>Provisional</td>
+                      <td>Parts of the visual image</td>
```
cconcolato, line 960:
```suggestion
                      <td>Indicates that the DAPT content represents any part of the visual of the programme</td>
```

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+                      <td>Audio Description</td>
+                      <td>See sub-types.</td>
+                    </tr>
+                    <tr>
+                      <td><code>visual;dialogue</code></td>
+                      <td>Provisional</td>
+                      <td>Verbal communication, in sign language</td>
```
cconcolato, line 967:
```suggestion
                      <td>Indicates that the DAPT content represents verbal communication, in sign language, within the visual of the programme</td>
```

index.html
```diff
@@ -995,7 +1118,11 @@ <h3>Script Event</h3>
               However, there can be cases where they do,
               e.g. in <a>Dubbing Scripts</a> when different <a>Characters</a> speak different text at the same time.</p>
           </li>
-          <li>An optional <a>Script Event Type</a> used to identify the type of event (e.g. dialogue, on screen text) in <a>Dubbing Scripts</a>.</li>
+          <li>A <a>Represents</a> property
+            used to identify what content the event represents (e.g. dialogue, on screen text, etc.).
+            Every <a>Script Event</a> MUST have a valid <a>Represents</a> property.
+          <aside class="note">The <a>Represents</a> property is inherited,
```
cconcolato, line 1124:
```suggestion
          <aside class="note">The <a>Represents</a> property may be inherited,
```

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+              <div class="registry-table-section">
+                <table class="data">
+                  <caption><a>Registry table</a> for the <code>&lt;<a>content-descriptor</a>&gt;</code> component
+                    whose Registry Definition is at <a href="#registry-table-content-descriptor"></a></caption>
+                  <thead>
+                    <th><code>&lt;<a>content-descriptor</a>&gt;</code></th>
+                    <th>Status</th>
+                    <th>Description</th>
+                    <th>Example usage</th>
+                    <th>Notes</th>
```
cconcolato, line 933: The notes column seems superfluous given the description and example usage.

index.html
cconcolato: And similar changes for the next rows...

examples/intro-script-with-gain.xml
```diff
@@ -1,10 +1,10 @@
 <tt ...
-  daptm:represents="visualNonText, visualText"
+  daptm:scriptRepresents="visual;nonText visual;text"
```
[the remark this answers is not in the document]
nigelmegitt: I'd prefer to explain, will do.

examples/intro-times-only.xml
```diff
@@ -1,5 +1,5 @@
 ...
-  <body>
+  <body daptm:represents="...">
```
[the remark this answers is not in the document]
nigelmegitt: I'll explain it

figures/sources/class-diagram.puml
```diff
@@ -13,7 +13,7 @@ skinparam Ranksep 80
 skinparam MinClassWidth 220
 
 Class DAPTScript as "**DAPT Script**" [[#dapt-script]] {
-    Represents [[[#represents]]]
+    Script Represents [[[#script-represents]]]
```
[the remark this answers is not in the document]
nigelmegitt: I'm happy to change the name. I also don't love the name Script Represents, but I'm slightly uncomfortable about all of those proposals! `documentRepresents` is probably the best option, but we tend to use "document" to refer to the TTML and "script" to refer to the DAPT Script. Hence the use of "Script Represents" in the data model.

index.html
```diff
@@ -885,6 +856,158 @@ <h5>Timing Properties</h5>
...
+            <div class="exampleInner">
+              <pre class="language-abnf nohighlight">
+    &lt;content-descriptor&gt;  # see registry table below
+    : &lt;descriptor-token&gt; ( &lt;descriptor-delimiter&gt; &lt;descriptor-token&gt; )*
+  
+    &lt;descriptor-token&gt;
+    : <a href="https://www.w3.org/TR/xmlschema-2/#NMTOKEN">xsd:NMtoken</a>
+
+    &lt;descriptor-delimiter&gt;
+    : ;  # SEMI-COLON U+003B
```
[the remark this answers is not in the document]
nigelmegitt: The issue with `.` is that it's permitted in URNs and in NMToken. I would be happy to prohibit URNs syntactically, choose a different reference definition than NMToken and use `.` for the delimiter.

I think it was your suggestion that we need to support URN @cconcolato , so perhaps you could consider if we really have that use case.

index.html
nigelmegitt: I think we need to keep the token list approach, to avoid accidental matching of partial words. We also need an implementable normative definition that's unambiguous.

We can certainly add that informative wording though.

index.html
nigelmegitt: Oh interesting, yeah I guess prefixed tokens might be what we need. Will think about that.

index.html
nigelmegitt: What is "the visual of the programme"?

nigelmegitt: Changing to "visual image of the programme"

index.html
nigelmegitt: Won't put a normative keyword in a note, but will change "inherited" to "inheritable".

index.html
nigelmegitt: I'd prefer to retain it.

index.html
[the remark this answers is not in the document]
cconcolato: Happy to not use URNs if we have a proper way of doing proprietary extensions.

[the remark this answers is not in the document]
nigelmegitt: OK will change to use `.` as delimiter and define it like NMToken but without the `.` being permitted.

index.html
nigelmegitt: Happy to allow new values to be permitted that have `x-` extension tokens appended to existing registry entries.

index.html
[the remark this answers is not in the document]
nigelmegitt: Done in 224440f

index.html
nigelmegitt: Done in 224440f

index.html
nigelmegitt: Done in 09b070e

index.html
nigelmegitt: Done in 09b070e

index.html
nigelmegitt: Done in 09b070e

index.html
nigelmegitt: Done in 09b070e

index.html
nigelmegitt: Done in 09b070e

index.html
nigelmegitt: Done in 09b070e

index.html
```diff
@@ -609,75 +616,35 @@ <h3>DAPT Script</h3>
         </p>
 
         <section>
-          <h4>Represents</h4>
-          <p>The <dfn>Represents</dfn> property is a mandatory property of a <a>DAPT Script</a> which
+          <h4>Script Represents</h4>
```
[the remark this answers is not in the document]
nigelmegitt: Done in 09b070e

examples/intro-times-only.xml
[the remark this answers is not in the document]
nigelmegitt: Done in 09b070e

examples/intro-times-and-text.xml
```diff
@@ -5,15 +5,15 @@
   ttp:contentProfiles="http://www.w3.org/ns/ttml/profile/dapt1.0/content"
   xml:lang="en"
   daptm:langSrc=""
-  daptm:represents="visualNonText, visualText"
+  daptm:scriptRepresents="visual;nonText visual;text"
```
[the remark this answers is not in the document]
nigelmegitt: Done in 09b070e

examples/intro-script-with-gain.xml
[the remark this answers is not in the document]
nigelmegitt: Done in 09b070e

index.html
```diff
@@ -844,11 +817,15 @@ <h4>Characters</h4>
...
 
         <section>
-          <h4>Shared properties</h4>
+          <h4>Shared properties and Value Sets</h4>
           <p>Some of the properties in the DAPT data model are common within more than one object type,
             and carry the same semantic everywhere they occur.
             These <dfn>shared properties</dfn> are listed in this section.
           </p>
+          <p>Some of the value sets in DAPT are reused across more than one property,
+            and have the same constraints everywhere they occur.
```
cconcolato, line 826: Represents and Script Represents don't really have put the same constraints on content-descriptor ... Maybe we should remove the end of the sentence.

cconcolato (approved):

index.html
nigelmegitt: There's no difference to the constraints on `<content-descriptor>`, there's a difference in how many content descriptors are permitted in Script Represents vs Represents, but that difference isn't described in this section.

PR #256, opened 2024-10-08
Make "Using computed attribute values" normative

### LINKED ISSUE #249: Non-Normative Section 6.4 contains normative keyword
>   Implementations MUST compute attribute values based on the contents of the 
document instance before applying those computed values to  DAPT data model objects.

This may be derived by TTML2 but it is still not defined as such in TTML2.

nigelmegitt:
Closes #249.

Retain and reformat the example about non-DAPT functionality.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/256.html" title="Last updated on Nov 13, 2024, 3:18 PM UTC (d0e3e3e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/256/e098bac...d0e3e3e.html" title="Last updated on Nov 13, 2024, 3:18 PM UTC (d0e3e3e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2548,9 +2548,8 @@ daptm:descType : string
         </aside>
       </section>
 
-      <section class="informative">
+      <section>
         <h3>Using computed attribute values</h3>
-        <p><em>Some normative provisions relating to this section are defined in [[TTML2]].</em></p>
         <p>Some attributes have semantics for computing their value that depend on the computed value of
           the attribute on some other element.
           For example if the <code>xml:lang</code> attribute is not specified on an element then its computed
@@ -2561,12 +2560,13 @@ daptm:descType : string
           of its descendant <code>&lt;div&gt;</code> elements that do correspond to <a>Script Events</a>
           are different to what they would be if that intermediate <code>&lt;div&gt;</code> element's <code>begin</code>
           time were ignored.</p>
-        <p class="ednote">MAYBE REMOVE Or, also for example, if styling is used in a TTML document, the computed value of the <code>tts:fontSize</code>
+        <aside class="example">An example of this relating to functionality not required by DAPT is styling.
+          If styling is used in a TTML document, the computed value of the <code>tts:fontSize</code>
           attribute when expressed as a percentage is relative to the computed value of the same attribute
           on the element's parent.
           Furthermore, in this example, the algorithm for generating ISDs (Intermediate Synchronic Documents)
           involves reparenting some elements. The computed value that applies is the one <em>before</em>
-          rehoming has taken place.</p>
+          rehoming has taken place.</aside>
         <p>Considering this situation more generally,
           it is possible that, within a DAPT document, there can be TTML elements that
           do not directly correspond to objects in the DAPT data model,
--- a/index.html
+++ b/index.html
@@ -2560,13 +2560,6 @@ daptm:descType : string
           of its descendant <code>&lt;div&gt;</code> elements that do correspond to <a>Script Events</a>
           are different to what they would be if that intermediate <code>&lt;div&gt;</code> element's <code>begin</code>
           time were ignored.</p>
-        <aside class="example">An example of this relating to functionality not required by DAPT is styling.
-          If styling is used in a TTML document, the computed value of the <code>tts:fontSize</code>
-          attribute when expressed as a percentage is relative to the computed value of the same attribute
-          on the element's parent.
-          Furthermore, in this example, the algorithm for generating ISDs (Intermediate Synchronic Documents)
-          involves reparenting some elements. The computed value that applies is the one <em>before</em>
-          rehoming has taken place.</aside>
         <p>Considering this situation more generally,
           it is possible that, within a DAPT document, there can be TTML elements that
           do not directly correspond to objects in the DAPT data model,
@@ -2576,6 +2569,9 @@ daptm:descType : string
         <p>The semantics defined by [[TTML2]] or, for vocabulary defined herein, this specification, take precedence in this scenario.
           Implementations MUST compute attribute values based on the contents of the <a>document instance</a> <em>before</em>
           applying those computed values to DAPT data model objects.
+          For example a processor that supports TTML2 styling features would need
+          to implement the TTML2 semantics for inheritance and computing relative values
+          of attributes like <code>tts:fontSize</code>.
         </p>
         <aside class="example">
           <p>This example demonstrates these possibilities,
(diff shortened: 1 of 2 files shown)


andreastai (approved):
Looks good to me! Thanks!

index.html
```diff
@@ -2554,12 +2553,13 @@ <h3>Using computed attribute values</h3>
           of its descendant <code>&lt;div&gt;</code> elements that do correspond to <a>Script Events</a>
           are different to what they would be if that intermediate <code>&lt;div&gt;</code> element's <code>begin</code>
```
cconcolato, line 2554: Reading this text again, I find the parts about "that element does not correspond to a Script Event" confusing. It shouldn't matter whether or not it is a Script Event. Inheritance applies. Also it gives the impression that under some circumstances, intermediate div can be ignored. 
I would remove the whole sentence starting with "If". Otherwise, I would rephrase saying something like: 
> If a workflow decides to rewrite a DAPT document to remove nested div elements, care should be taken that if those elements had inheritable attributes like begin, the attribute values like begin on the nested divs should be adjusted.

index.html
```diff
@@ -2554,12 +2553,13 @@ <h3>Using computed attribute values</h3>
...
           time were ignored.</p>
-        <p class="ednote">MAYBE REMOVE Or, also for example, if styling is used in a TTML document, the computed value of the <code>tts:fontSize</code>
+        <aside class="example">An example of this relating to functionality not required by DAPT is styling.
+          If styling is used in a TTML document, the computed value of the <code>tts:fontSize</code>
           attribute when expressed as a percentage is relative to the computed value of the same attribute
           on the element's parent.
           Furthermore, in this example, the algorithm for generating ISDs (Intermediate Synchronic Documents)
           involves reparenting some elements. The computed value that applies is the one <em>before</em>
-          rehoming has taken place.</p>
+          rehoming has taken place.</aside>
```
cconcolato, line 2562: Do we really need this example? It is about a "functionality not required by DAPT". I would remove it altogether. If not, I would rephrase saying something like:
> If a DAPT document contains additional features not required by DAPT such as styling and the DAPT document is processed by a processor supporting these features, support for inheritance of the inheritable features such as tts:fontSize, is required, per TTML2.

index.html
nigelmegitt: I don't like the idea of making this example specifically about rewriting documents - it's about the correct interpretation of a document, regardless of what is being done with it.

Possibly this comment should be raised as a separate issue?

I'd be happy to emphasise further (informatively) that ignoring the intermediate `<div>` is not an acceptable thing to do.

index.html
nigelmegitt: Removing it was one of the options before, but until nobody came down on the side of "remove". Not wanting to leave an editorial note in, if avoidable, I just turned it into an example. I don't mind deleting it if it isn't helpful.

index.html
cconcolato: My suggestion was a tentative rewrite of what I thought you were trying say. The alternative was to remove the sentence. My main problem is the current emphasis on elements that do not correspond to Script Events. We should not encourage that.

index.html
cconcolato: Is my rephrasing not acceptable?

nigelmegitt: Sorry, I understood your preference to be for removal, and didn't focus on the rephrasing. Reading it again, I think the following two paragraphs already say the same thing as the rephrased version - maybe we should just add to one of them the example of `tts:fontSize` to make the abstract statements more concrete, and remove the example?

Concretely, I'm suggesting the paragraph:

>The semantics defined by [[TTML2](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/256.html#bib-ttml2)] or, for vocabulary defined herein, this specification, take precedence in this scenario. Implementations MUST compute attribute values based on the contents of the [document instance](https://www.w3.org/TR/ttml2/#terms-document-instance) before applying those computed values to DAPT data model objects.

would become:

>The semantics defined by [[TTML2](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/256.html#bib-ttml2)] or, for vocabulary defined herein, this specification, take precedence in this scenario. Implementations MUST compute attribute values based on the contents of the [document instance](https://www.w3.org/TR/ttml2/#terms-document-instance) before applying those computed values to DAPT data model objects. **For example a processor that supports TTML2 styling features would need to implement the TTML2 semantics for inheritance and computing relative values of attributes like `tts:fontSize`.**

cconcolato: and remove the example?

nigelmegitt: Yes, and remove the example.

index.html
nigelmegitt: Here's another attempt at what I was trying to say, maybe this is less confusing?

>Or, for another example, the computed times of an element in a DAPT document are relative to the begin time of its parent element. If a `<div>` element specifies a `begin` attribute, then the computed times of its child `<div>` elements are relative to that intermediate `<div>` element's begin time, and so on down the hierarchy. It is important to include those intermediate `<div>` elements' times in the computation even if the processing target is an instance of the DAPT data model in which they have no direct equivalent; otherwise the [Script Event](https://www.w3.org/TR/dapt/#dfn-script-event) Begin and End times would be wrong.

index.html
nigelmegitt: Done in 5617ab0.

index.html
nigelmegitt: Change pushed in 33c33fe. On re-reading, I made a couple of small tweaks because I thought the concept of an "intermediate" div wasn't really explained properly.

cconcolato (approved):
Overall, I'm approving the PR to unlock the CR. 

That said, I don't think it is a substantive change because the changes are about examples. The section contains a single MUST statement that is not changed.

Anyway, I think we could improve the section editorially later on. I find it scary for implementers because the boundaries of the implementation requirements are not clear. There are many examples (xml:lang, begin with nested divs, styles), but the concrete normative text is not visible enough, upfront. As I understand it, there are only 2 concrete consequences on implementers:
* use of style TTML2 properties, and there are 2 cases: either the implementation cares about them and it needs to implement proper inheritance; or it does not care about them and can fully ignore them
* use of nested divs (which I'm not a big fan of...). Implementations don't have a choice. To be compliant, they have to apply the inheritance model on intermediate divs.

nigelmegitt:
> That said, I don't think it is a substantive change because the changes are about examples. The section contains a single MUST statement that is not changed.

Thanks @cconcolato - the reason the section was made normative was because it had a MUST statement in, but was previously marked as non-normative.

> there are only 2 concrete consequences on implementers

Not just style properties and nested divs, also timing properties and anything else that can be put on a div, like `xml:lang`.

Thanks for the approval, I will merge, and if more problems or improvements are identified they can be dealt with in new issues and pull requests.

PR #265, opened 2024-10-09
Check feature dispositions

### LINKED ISSUE #264: Align section F.1 (Disposition of features) with section F.2 (DAPT Content profile)
All optional features in the DAPT content profile must be listed in section F1 as permitted. At the moment some are missing (e.g. #unicodeBidi).

nigelmegitt:
Closes #264 by adding missing feature dispositions. Also adds term definition links to each TTML2 feature.


nigelmegitt:
I noticed that the default Respec behaviour is to use the TTML2 1st Ed Rec, but we are referencing TTML2 features that are only introduced in 2nd Edition, so I updated the pull request to reference the most recent TTML2 CR Snapshot - unfortunately even though SpecRef knows that this is the most recent version of `ttml2` it doesn't reference it. The alternative is to be explicit and point to [`[ttml2-20210309]`](https://www.specref.org/?q=ttml2-20210309) but that's quite long winded, so instead I added a local bibliography reference.

Happy to revert it if we know a better way to alias `[TTML2]` to the latest 2nd Ed version.

andreastai:
To be consistent it should say for #profile-root: "This is the profile expression of the prohibition of [5.6.3 ttp:profile](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/265.html#ttp-profile)."

nigelmegitt:
> To be consistent it should say for #profile-root: "This is the profile expression of the prohibition of [5.6.3 ttp:profile](https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/265.html#ttp-profile)."

@andreastai Done (slightly differently) in 2bf8b2c:

>This is the profile expression of the prohibition of the `ttp:profile` attribute on the `<tt>` element as specified in 5.6.3 ttp:profile.

andreastai (changes requested):
*  feature `#contentProfiles` is required in dapt processor profiles but it is optional in the dispositions
* extension `represents` is required in the dapt content profile but it is "only" permitted in the dispositions

Thanks @nigelmegitt 

Otherwise the changes look good to me.

nigelmegitt:
Thanks @andreastai , good catches! Fixed in 7dc454b.

andreastai (approved):
Thanks. Fixes in https://github.com/w3c/dapt/commit/7dc454ba9a60e8c6c542b8f257a3b683fa9d1166 look good to me.

PR #273, opened 2024-11-15
Add an XSD

### LINKED ISSUE #272: Add an XSD for simple validation
So that people can do basic validation of DAPT documents.

nigelmegitt:
Closes #272 by adding an XSD based on the TTML2 XSD (by copy rather than reference) and making additions and modifications as needed to meet the constraints of DAPT.

With thanks to @magicbadger for his help and content while developing this.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/273.html" title="Last updated on Mar 21, 2025, 2:08 PM UTC (0d5b540)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/273/b688115...0d5b540.html" title="Last updated on Mar 21, 2025, 2:08 PM UTC (0d5b540)">Diff</a>


himorin:
just note, all http://www.w3.org/ns are automatically redirected to https
might not cause error over tooling (per past survey on whole ns redirection), but leave comment here just to note.

nigelmegitt:
> just note, all http://www.w3.org/ns are automatically redirected to https might not cause error over tooling (per past survey on whole ns redirection), but leave comment here just to note.

Are you suggesting that the Namespaces in the spec should be changed to `https` (which would be a separate issue)? I don't think there's any requirement that anyone or any implementation actually fetches the resource at the URL, and certainly I don't expect the XSD to cause that. I'd just like to know more about what concerns there might be about the https redirect.

cconcolato:
@nigelmegitt we have been testing with some XML output of our TTAL converter and getting some errors like:
```
file.xml:33: element layout: Schemas validity error : Element '{http://www.w3.org/ns/ttml}layout': This element is not expected.
file.xml:41: element div: Schemas validity error : Element '{http://www.w3.org/ns/ttml}div', attribute '{http://www.w3.org/ns/ttml/profile/dapt#metadata}onScreen': The attribute '{http://www.w3.org/ns/ttml/profile/dapt#metadata}onScreen' is not allowed.
file.xml:41: element div: Schemas validity error : Element '{http://www.w3.org/ns/ttml}div', attribute 'region': The attribute 'region' is not allowed.
```
My memory of XSD is blurry but looking at the schemas, I don't see where onScreen is explicitly allowed on `div` for example.

nigelmegitt:
Thanks @cconcolato , looks like I was too keen on stripping out TTML2 region binding that we didn't mean to prohibit, so I have restored that in ff2d5ca. Also explicitly added `daptm:onScreen` to `<div>`.

I can't reproduce the unexpected layout element error, so if that persists for you, please could you share the document that's failing validation?

nigelmegitt:
Agenda+ for 2025-12-19 TTWG call to see if anyone has any recommendations for validation tools or approaches using the XSD

PR #286, opened 2025-03-07
Fix darkmode presentation

nigelmegitt:
Dark mode somehow got broken in the last 7 months since I last fixed it.

The current toggles don't fix the media query for prefers darkmode, but just add the `darkmode` class to the `body` element. Explicitly set the colours when the `darkmode` class is added to `body`, horribly duplicating the existing `@media (prefers-color-scheme: dark)` rule but using CSS nesting.

This is a hack but does make all the text at least readable in dark mode.


nigelmegitt:
@himorin I suggest we merge this and also merge it into the CR branch as soon as you can approve it, because the current CR version is broken now in dark mode. 

Actually it's _still_ broken but not quite as badly - the dark mode toggle tries to remember the setting but doesn't re-apply it on page load, you have to explicitly click it to Light or Auto and then back to Dark to get the rest of the page to load properly, but that's a problem with someone else's script - probably fixup.js, I imagine.

himorin:
if we want to introduce duplicate CSS definition from prefers-color-scheme to .darkmode, I suppose we should do the same thing for light with default mode as darkmode..
I believe it could be performed with `@media (prefers-color-scheme: light)` with `:not(:has(.darkmode))`, but haven't checked yet.

nigelmegitt:
> if we want to introduce duplicate CSS definition from prefers-color-scheme to .darkmode, I suppose we should do the same thing for light with default mode as darkmode.. I believe it could be performed with `@media (prefers-color-scheme: light)` with `:not(:has(.darkmode))`, but haven't checked yet.

I don't think we need to - it's the default behaviour and seems to work fine without doing that.

nigelmegitt:
> I suppose we should do the same thing for light with default mode as darkmode..

@himorin just to add, if you have any reproducible steps that show a problem with the light mode styling not being applied in light mode, please tell me! I tried all the ways I could think of and could not create a problem with the CSS as it is in this PR.

In Firefox's dev tools it shows that the `@media (prefers-color-scheme: light) {}` block rules are being applied in light mode.

nigelmegitt:
@himorin I think that's fixed now: are you able to check and confirm?

gkatsev:
I can reproduce that, but it seems to only happen when the browser is in light mode and choosing dark mode on the page toggle.

nigelmegitt:
It's also worth noting that what that sidebar toggle does is just adds or removes the `darkmode` class on the `<body>` independently of the browser's dark mode setting. So there are permutations of the media query result and the `darkmode` class being present or absent - in short, I'm not at all surprised that you found 4 distinct presentations, and I don't think it's worth trying to fix for all of them, because it looks like a broken pattern to me.

Having said that, this PR does set the styles based on both the media query and the `darkmode` class so I'm a bit disappointed that it still has problems.

gkatsev:
ah, can trigger it also by selecting dark mode and using the extension to toggle the browser darkmode setting. Guess it's a mismatch between respec's `prefer-color-scheme: dark` and `.darkmode`.
It's probably possible to fix, but maybe not a blocker for this PR

nigelmegitt:
> The full solution likely involves dealing with the media query and the css class and fixing each permutation

Also fixing the W3C stylesheets to do the same.

nigelmegitt:
> It's probably possible to fix, but maybe not a blocker for this PR

Not only not a blocker for this PR but also not a blocker for CR publication, assuming this PR gets merged.

nigelmegitt:
Are you able to get this into the CR for tomorrow, @himorin ?

himorin (approved):

Issue #295, opened 2025-06-10
Should we allow `Represents` on `Text` objects?


nigelmegitt:
I don't think we currently allow `Represents` on `Text` objects (that would be represented in XML with the `span` element). We would need to change: a) the general semantics to say that it applies to Script Event and Text and b) the Class diagram.

_Originally posted by @cconcolato in https://github.com/w3c/dapt/pull/294#discussion_r2129157574_

PR #300, opened 2025-06-23
Explicitly permit daptm:represents on tt, body, div, p and span

### LINKED ISSUE #262: Clarify usage of Represents and Script Represents
Section 4.7 states

> The [Represents](https://w3c.github.io/dapt/#dfn-represents) property is inheritable. If it is absent from an element then its computed value is the computed value of the [Represents](https://w3c.github.io/dapt/#dfn-represents) property on its parent element, or, if it has no parent element, it is the empty string. If it is present on an element then its computed value is the value specified.

Currently it is not possible that a parent element can have a `@represents` attribute as script events can not be nested and `@represents` can only be used on script events.

> Note
Since there is no empty <[content-descriptor](https://w3c.github.io/dapt/#dfn-content-descriptor)>, this implies that an empty computed [Represents](https://w3c.github.io/dapt/#dfn-represents) property can never be valid; one way to construct a valid [DAPT Document](https://w3c.github.io/dapt/#dfn-dapt-document) is to specify a [Represents](https://w3c.github.io/dapt/#dfn-represents) property on the [DAPT Script](https://w3c.github.io/dapt/#dfn-dapt-script) so that it is inherited by all descendants that do not have a [Represents](https://w3c.github.io/dapt/#dfn-represents) property.

Only Script Represents can be specified on the DAPT Script not Represents (at least for now).

nigelmegitt:
Closes #262 and #295.

Explicitly state the previous implied permission to put daptm:represents on content elements including `<p>` and `<span>`.

Update the XSD to reflect this permission (doesn't make any practical difference since 'other' metadata was already permitted, but it is clearer to see).

Add the attribute to each of the elements in the schema validator test fixture to verify that it continues to permit it and passes the test.

Built on #294 (without the commit that removed permission for represents on `<p>` and `<span>`).

Adds Represents to Text, both in the class diagram and in the data model descriptions of Text and Represents.

Includes a note advising that where possible, script events should be split if they represent different things.


index.html
```diff
@@ -1429,6 +1429,27 @@ <h4>Text</h4>
...
+            <a>Text</a> objects in different <a>Script Events</a> each with their own <a>Represents</a> property.
+            The option to assign <a>Represents</a> directly to <a>Text</a> objects or parts of <a>Text</a> objects
+            is provided in case this is impractical.</p>
+            <p>For example, consider a <a>Script Event</a> authored to describe both
+              non-textual parts of the video image and
+              visual text content, that is intended to be voiced in a single pass.
+              If this were split into two or more <a>Script Events</a> then it would be necessary
+              to predict with audio sample accuracy when the recording will switch between
+              the each <a>Script Event</a>, in order to get correct times.
+              Doing would likely lead to timing errors, though the times could be adjusted after the recording.
```
cconcolato, line 1446: I know it's a note but the notion of "predicting duration with sample accuracy" or "timing errors" feels too detailed and unnecessary.
 ```suggestion
              If this were split into two or more <a>Script Events</a> then it would be necessary
              to assign an end time to the first one and a start time to the second one, 
              which may not be simple.
```

File: index.html [hunk 2]
```diff
@@ -1429,6 +1429,27 @@ <h4>Text</h4>
           <p>The source language of <a>Translation</a> <a>Text</a> objects and, where applicable,
             <a>Original</a> <a>Text</a> objects
             is indicated using the <a>Text Language Source</a> property.</p>
+          <p>If the <a>Text</a> object, or part of it, represents something more specific, or different,
+            to its parent <a>Script Event</a>, it can have a different <a>Represents</a> property.</p>
```
cconcolato, line 1433: I don't think we should encourage the Text object itself to have a different "represents" than the parent. 
```suggestion
          <p>If a part of a <a>Text</a> object represents something more specific, or different,
            to its parent <a>Script Event</a>, it can have a different <a>Represents</a> property.</p>
```

index.html
```diff
@@ -2728,6 +2766,95 @@ <h3>Using computed attribute values</h3>
...
+            <tr>
+              <td><code>end</code></td>
+              <td><a>End</a></td>
+              <td><code>&lt;body&gt;</code>,
+                <code>&lt;div&gt;</code>,
+                <code>&lt;p&gt;</code>,
+                <code>&lt;span&gt;</code>,
+                <code>&lt;audio&gt;</code>,
+                <code>&lt;animate&gt;</code></td>
+              <td>Constrains the computed end time of its descendants</td>
```
cconcolato, line 2818:
```suggestion
              <td>Constrains the computed end time of its descendants, as defined in [[TTML2]]</td>
```

index.html
```diff
@@ -2728,6 +2766,95 @@ <h3>Using computed attribute values</h3>
...
+            </tr>
+            <tr>
+              <td><code>daptm:langSrc</code></td>
+              <td><a>Text Language Source</a></td>
+              <td><code>&lt;tt&gt;</code>,
+                <code>&lt;body&gt;</code>,
+                <code>&lt;div&gt;</code>,
+                <code>&lt;p&gt;</code>,
+                <code>&lt;span&gt;</code></td>
+              <td>Inherited</td>
```
cconcolato, line 2839:
```suggestion
              <td>Inherited, as defined in this section</td>
```

schema-validator/test/fixtures/valid_dapt.ttml
```diff
@@ -24,11 +25,15 @@
...
     </head>
-    <body>
+    <body daptm:represents="visual.nonText">
         <div xml:id="se1" begin="3s" end="10s" ttm:agent="character_2" daptm:represents="audio.dialogue" daptm:onScreen="ON">
             <ttm:desc daptm:descType="scene">high mountain valley</ttm:desc>
             <metadata></metadata>
             <p daptm:langSrc="en"><span>Look at this beautiful valley.</span></p>
         </div>
+        <div xml:id="se2" begin="13s" end="20s" daptm:represents="visual.nonText">
+            <metadata></metadata>
```
cconcolato, line 35: What is the value of this empty metadata element?

index.html
cconcolato: Sorry for the confusion. There are 2 separate topics:
1. I think we should create a new issue to mention the case "if the semantic structure of the script is to have multiple paragraphs in the same Script Event". That is not obvious to me.
2. I think my initial suggestion becomes irrelevant.

PR #313, opened 2025-09-05
Use xref to resolve TTML2 and w3c-process links

### LINKED ISSUE #308: Links to TTML2 terms/sections/etc do not all resolve to the same version of TTML2
Referring to:
https://www.w3.org/TR/2025/CRD-dapt-20250801/

Some of the links to terms/sections etc begin:
`https://www.w3.org/TR/2021/cr-ttml2-20210309#`
whereas others begin:
`https://www.w3.org/TR/ttml2/#`

These URLs resolve to different editions of TTML2 (see https://github.com/w3c/imsc/issues/612 for relevant background info)

nigelmegitt:
Fixes #308.


nigelmegitt:
I agree, removing TTML2 from `xref` makes no difference to the generated output, so I will remove it. I'm a bit surprised by that, from the documentation, but "less is more" here.

PR #314, opened 2025-09-11
Match TTML2 in what elements a src can reference

### LINKED ISSUE #312: Use of src on embedded content entities appears to conflict with TTML2
https://www.w3.org/TR/dapt/#audio-recording includes the following (and related examples):
  
> * If the [Source](https://www.w3.org/TR/dapt/#dfn-source) is an embedded audio resource, the [Source](https://www.w3.org/TR/dapt/#dfn-source) and [Type](https://www.w3.org/TR/dapt/#dfn-type) properties are represented together by exactly one of:
> 1. A src attribute that is a fragment identifier that references **either an [\<audio\>](https://www.w3.org/TR/ttml2/#embedded-content-vocabulary-audio) element or a [\<data\>](https://www.w3.org/TR/ttml2/#embedded-content-vocabulary-data) element**, where the referenced element is a child of /tt/head/resources and specifies a type attribute and the xml:id attribute used to reference it;
>   This mechanism cannot be used if there is more than one [Source](https://www.w3.org/TR/dapt/#dfn-source).
>
> 2. A [\<source\>](https://www.w3.org/TR/ttml2/#embedded-content-vocabulary-source) child element with a src attribute that is a fragment identifier that references **either an \<audio\> element or a [\<data\>](https://www.w3.org/TR/ttml2/#embedded-content-vocabulary-data) element**, where the referenced element is a child of /tt/head/resources and specifies a type attribute and the xml:id attribute used to reference it;
> 3. ...

Yet this appears to conflict with https://www.w3.org/TR/2021/CR-ttml2-20210309/ which requires the src attribute of an audio entity to be an audio value expression:

> If an \<audio\> value expression takes the form of a bare fragment identifier, then that fragment identifier **must reference an audio element that is a child of a resources element** in the enclosing document instance.

and requires the src attribute of a source entity to be a data value expression:

> If a \<data\> value expression takes the form of a fragment identifier, then that fragment identifier **must reference a data element that is a child of a resources element** in the enclosing document instance.

It looks like this is related to #220 (and perhaps others)

nigelmegitt:
Fixes #312 by making it that an `<audio>` element's `src` attribute must reference an `<audio>` element and a `<source>` element's `src` attribute must reference a `<data>` element.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/314.html" title="Last updated on Sep 23, 2025, 5:32 PM UTC (b330956)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/314/8836201...b330956.html" title="Last updated on Sep 23, 2025, 5:32 PM UTC (b330956)">Diff</a>

### DIFF
--- a/substantive-changes-summary.txt
+++ b/substantive-changes-summary.txt
@@ -76,3 +76,5 @@ From CRS (20250311)
 * Allow Represents to apply to Text (#295)
 
 * Replace the `#scriptEventGrouping` extension feature with the TTML1 feature `#nested-div` (#303)
+
+* Clarify which elements the `src` attribute can reference using a fragment identifier, and align with TTML2 (#312)


thomasheritage:
Thanks @nigelmegitt -- yes, it looks like this resolves #312

nigelmegitt:
> Thanks @nigelmegitt -- yes, it looks like this resolves #312

@thomasheritage many thanks!

PR #328, opened 2025-10-17
Clarify requirements for Represents

### LINKED ISSUE #327: Tidy up language around Represents
While implementing validation for Represents I noticed that the requirement for a _computed_ value of Represents on Script Event and Text objects is unclear. The wording under Script Event and Text is slightly different, and it is not obvious why.

It also is unclear if it is acceptable to have a `<span>` that sets `daptm:represents=""` if its ancestor `<p>` has a computed value of `daptm:represents` that is valid.

Suggest tidying the language to make this less ambiguous:

* Clarify that every part of a Text must have a valid computed Represents
* Clarify that Script Events and Text objects must have a valid computed Represents
* Clarify that `<div>` elements that are not Script Events don't need a valid computed Represents

It might be useful to specify in [§4.7 Represents](https://www.w3.org/TR/dapt/#represents) which objects the property _applies_ to, as well as which is may be specified on, which is listed now.

nigelmegitt:
* Require valid computed valid for Represents in Text object and all its sub-parts
* Add definition wrappers to namespace prefixes
* Define locally, and exclude, metadata properties of Script Events when deciding if it's possible to map from a `<div>` to a Script Event.

Closes #327.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/328.html" title="Last updated on Nov 26, 2025, 10:07 AM UTC (313066c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/328/1d92778...313066c.html" title="Last updated on Nov 26, 2025, 10:07 AM UTC (313066c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1406,6 +1406,12 @@
             is indicated using the <a>Text Language Source</a> property.</p>
           <p>If the <a>Text</a> object, or part of it, represents something more specific, or different,
             to its parent <a>Script Event</a>, it can have a different <a>Represents</a> property.</p>
+          <p>The <a>Text</a> object, and every one of its parts, MUST have a valid <a>Represents</a> property.
+            The presence of the property itself is considered optional
+            because if the <a>Text</a> object omits the
+            property then the property is inherited from the parent <a>Script Event</a>.
+            However it is an error for a <a>Text</a> object to specify an invalid <a>Represents</a> property.
+          </p>
           <aside class="note">
             <p>Where possible, text representing
             different parts of the <a>related media object</a> should be put into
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1406,6 +1406,12 @@ <h4>Text</h4>
             is indicated using the <a>Text Language Source</a> property.</p>
           <p>If the <a>Text</a> object, or part of it, represents something more specific, or different,
             to its parent <a>Script Event</a>, it can have a different <a>Represents</a> property.</p>
+          <p>The <a>Text</a> object, and every one of its parts, MUST have a valid <a>Represents</a> property.
+            The presence of the property itself is considered optional
+            because if the <a>Text</a> object omits the
+            property then the property is inherited from the parent <a>Script Event</a>.
+            However it is an error for a <a>Text</a> object to specify an invalid <a>Represents</a> property.
```
cconcolato: This sentence seems unnecessary.

index.html
```diff
@@ -1406,6 +1406,12 @@ <h4>Text</h4>
             is indicated using the <a>Text Language Source</a> property.</p>
           <p>If the <a>Text</a> object, or part of it, represents something more specific, or different,
             to its parent <a>Script Event</a>, it can have a different <a>Represents</a> property.</p>
+          <p>The <a>Text</a> object, and every one of its parts, MUST have a valid <a>Represents</a> property.
+            The presence of the property itself is considered optional
+            because if the <a>Text</a> object omits the
+            property then the property is inherited from the parent <a>Script Event</a>.
```
cconcolato: I'm not sure I understand why we need these additional sentences. 
Regarding the first sentence, any use of any property must use a valid value. 
Regarding the second, it is already covered by the notion of the computed value.

nigelmegitt: The text does not mention the computed value though.

The reason I added this text is to ensure that it is clear that sub-parts of a Text need to have a valid `daptm:represents`.

For example:

```xml
<p daptm:represents="x-valid">This is okay</p>
```

but is this okay?

```xml
<p daptm:represents="x-valid">Text with a <span daptm:represents="invalid">sub-part</span> that needs to be not okay</p>
```

In the second example, the Text (the outer `<p>`) has a valid `daptm:represents` but also contains a `<span>` that does not. Without this additional spec text I think it is not clear if that is valid.

cconcolato: ok

cconcolato (approved):

PR #333, opened 2025-12-08
Add section about IMSC compatibility

### LINKED ISSUE #332: Add note about compatibility with IMSC
See also w3c/imsc#622.

Add a new informative appendix about "Compatibility with other TTML-based specifications" and a subsection on IMSC.

Broadly, a conformant DAPT document is also a conformant IMSC Text document, as long as it meets the constraints of both specifications. Where processing by an IMSC processor is desirable, the IMSC Text profile designator can be included in `ttp:contentProfiles` as an additional value to the DAPT profile designator.

It may be worth mentioning that:

* adding layout and styling to a DAPT document may be a useful way to produce a subtitle or caption file
* the computed values of `xml:lang` and `daptm:srcLang` may be useful to derive appropriate values for `itts:forcedDisplay`
* in some applications it may be better to map a single DAPT document into multiple IMSC Text documents, each intended for a specific single language. This can be done by pruning the content with an unwanted language, for example.
* in some applications it may be better to process the text content of a DAPT document to block it out into presented lines or blocks of lines of text in distinct ISDs, for readability.

nigelmegitt:
Addresses #332.

* Export some terms so they can be used in other specs e.g. IMSC
* Add a section on signalling conformance to IMSC
* Add a section on IMSC compatibility, including a subsection on how DAPT metadata can be used in the subtitle production workflow.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/pull/333.html" title="Last updated on Jan 19, 2026, 3:05 PM UTC (52a6488)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/dapt/333/e50c715...52a6488.html" title="Last updated on Jan 19, 2026, 3:05 PM UTC (52a6488)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2500,7 +2500,22 @@ daptm:descType : string
           </ul>
         </section>
 
+        <section id="imsc-signaling">
+          <h4>Signaling conformance to IMSC</h4>
+
+          <p>A document conforming to this specification and to [[ttml-imsc]]
+            signals its conformance to both by
+            including in the <code>ttp:contentProfiles</code> attribute
+            both the designators for <a>DAPT 1.0 Content Profile</a> and
+            <!-- [=IMSC Text Profile=] -->
+            <a data-cite="ttml-imsc1.3#dfn-text-profile">IMSC Text Profile</a>,
+            in addition to any other profiles to which the document conforms.
+          </p>
+
+          <p class="note">See also <a href="#imsc-interop"></a>.</p>
+        </section>
       </section>
+
       <section>
         <h3>Timing constraints</h3>
         <p>Within a DAPT Script, the following constraints apply in relation to time attributes and time expressions:</p>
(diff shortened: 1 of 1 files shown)


nigelmegitt (commented):
Self review:

* Somehow this introduces DAPT as a reference
* It also ends up referencing three different versions of IMSC
* The IMSC profile signalling section should be moved into the new appendix, rather than where it is at the time of this review, in the DAPT profile signalling section.

index.html
```diff
@@ -4501,6 +4517,86 @@ <h3><code>&lt;content-descriptor&gt;</code> registry table definition</h3>
...
+    <h2>Compatibility with other TTML-based specifications</h2>
+
+    <section class='appendix'>
+      <h3>Overview</h3>
+
+      <p>This specification is designed to be compatible with [[ttml-imsc1]]
+        and other profiles of [[ttml2]].
+        Specifically, by selecting a subset of the features and
+        extensions defined in this specification,
+        it is possible to create a document that:conforms to [[ttml-imsc1]],
```
palemieux:
```suggestion
        it is possible to create a document that conforms to [[ttml-imsc1]],
```

palemieux (approved):
I am not in love with duplicating prose between DAPT and IMSC 1.3.

Ideally the compatibility prose would be in a single document.

Perhaps it makes most sense to have it only in the DAPT document since presumably making DAPT docs conform to IMSC helps DAPT because there are already IMSC renderers?

If the prose is in the DAPT doc, then the IMSC doc can informatively point to it.

index.html
nigelmegitt: done in 262ad0c

nigelmegitt:
> Perhaps it makes most sense to have it only in the DAPT document since presumably making DAPT docs conform to IMSC helps DAPT because there are already IMSC renderers?

I don't think this is the motivation. Rather, it is because it makes sense in some workflows to use DAPT as a starting point to make subtitles or captions, which would presumably be IMSC documents, and it is useful to note the potential benefits of that approach and how it can be done.

> I am not in love with duplicating prose between DAPT and IMSC 1.3.
> 
> Ideally the compatibility prose would be in a single document.

...

> If the prose is in the DAPT doc, then the IMSC doc can informatively point to it.

Hmm, this is worth discussing. I duplicated it because I think it is useful and important for readers of both specifications, and it's generally preferable not to make people click through to other documents unless that's strictly needed. I'm happy to be flexible with this, but I can't work out which is the better spec to contain the text, and which should refer to it, if we take that approach!

I guess the first point to consider is if the text should actually be the same, or if there are some aspects that are more important for readers of IMSC and others for readers of DAPT.

palemieux:
> I don't think this is the motivation. Rather, it is because it makes sense in some workflows to use DAPT as a starting point to make subtitles or captions, which would presumably be IMSC documents, and it is useful to note the potential benefits of that approach and how it can be done.

Having both DAPT and IMSC  be subset of TTML is greatly useful since concepts and software building blocks can be reused.

In what scenario(s), would having a document conform to both DAPT and IMSC be useful if an IMSC document (created for distribution) cannot generally be used in a DAPT workflow, and vice versa?

nigelmegitt:
> In what scenario(s), would having a document conform to both DAPT and IMSC be useful if an IMSC document (created for distribution) cannot generally be used in a DAPT workflow, and vice versa?

Adding DAPT metadata to an IMSC document intended for client playback could be very useful, and even drive player behaviour in "advanced" players, for example by offering some method for viewing the speaker names that might not be in the text being displayed in a particular ISD, or for driving speech to text for translation subtitles mixed in with non-translation subtitles.

The features of IMSC Text that are absent from DAPT are essentially styling and layout. Adding either of those to a DAPT document could be used in some authoring environment to affect presentation, e.g. changing how text looks to provide emphasis in the text, that might affect how an actor voices the text when making an audio recording. However I'd caution that a DAPT document that includes translations into multiple languages might make no sense to use in the context of IMSC using an IMSC native player.

In summary:

* Using an IMSC document in a DAPT workflow _could_ be possible, but it's not clear to me _why_ someone would do that, maybe using hard of hearing subtitles as a starting point for a transcript of the audio, I suppose, prior to translation (my expectation is to begin with a clean transcript, more commonly, but I can imagine skipping that for cost/pragmatism reasons);
* Using a DAPT document in an IMSC workflow would clearly make sense, though I suspect that it would most likely make sense to generate _n_ DAPT+IMSC documents from _1_ DAPT document if that starting document contains multiple language variants;
* In either case it's not necessarily the case that a document created to conform to one profile can _generally_ be used in a workflow intended for the other, but there are likely some good benefits in _specific_ use cases.

palemieux:
@nigelmegitt In light of https://github.com/w3c/dapt/pull/333#issuecomment-3769483232, in what circumstance(s) would signaling that a document conforming to both DAPT and IMSC be useful? IMSC renderers will not do anything with the DAPT metadata, and the IMSC document is likely missing information to be useful in DAPT workflows.

In order to determine in what location(s) the prose would be useful, I am encouraging us to figure out in what scenario(s) having a document conforming to both IMSC and DAPT  would be used.

nigelmegitt:
> @nigelmegitt In light of [#333 (comment)](https://github.com/w3c/dapt/pull/333#issuecomment-3769483232), in what circumstance(s) would signaling that a document conforming to both DAPT and IMSC be useful? IMSC renderers will not do anything with the DAPT metadata, and the IMSC document is likely missing information to be useful in DAPT workflows.
> 
> In order to determine in what location(s) the prose would be useful, I am encouraging us to figure out in what scenario(s) having a document conforming to both IMSC and DAPT would be used.

Right, good question. Some reasons I can think of:

1. to allow an IMSC processor to adjust its behaviour to offer additional presentation features if DAPT metadata is present, potentially supporting features beyond the specification;
a. an example would be a preprocessor whose input has IMSC styling applied to DAPT content, and whose output is a set of IMSC documents any one of which a conformant IMSC processor would present appropriately, by using the DAPT metadata to select or modify the content - e.g. to include only dialogue subtitles (omitting sound effects) or to modify the indication of change of speaker;
b. another example would be a presentation processor that can provide to the hosting page a list of character names (and maybe actor names) speaking in the current scene, extracting the information from the DAPT metadata;
3. if a presentation processor can _only_ present documents that conform to both then it would presumably need both to be signalled;
4. to allow a processor that can present either one or the other profile to know that it can process the document that signals conformance to both.

Does this help with the question of where the prose would be useful? I'm struggling to think of a strong reason why it would be better in one place than the other.

palemieux:
> 1. to allow an IMSC processor to adjust its behaviour to offer additional presentation features if DAPT metadata is present, potentially supporting features beyond the specification; [...]

Perhaps, but this is no longer an IMSC processor as defined in IMSC 1.3.

We could consider adding additional semantic vocabulary to IMSC, but that is a longer discussion, beyond IMSC 1.3, beyond merely signaling conformance to DAPT and one that requires clear use cases.

> if a presentation processor can only present documents that conform to both then it would presumably need both to be signalled;

I do not see a clear use case for this. Why would a processor be limited in such a way?

> to allow a processor that can present either one or the other profile to know that it can process the document that signals conformance to both.

Sure. This is already possible without elaboration since conformance signaling in IMSC and DAPT are not mutually exclusive. It is however still not clear to me why someone would want (in practice) a document that conforms to both IMSC 1.3 and DAPT.

Per our offline discussion, it might be useful to provide guidance in DAPT how to create DAPT documents that can be previewed in an IMSC 1.3 player, i.e., here are the constraints and pitfalls of doing so.

PR #336, opened 2026-04-07
updating action used in ghaction

himorin:
to the latest one
https://github.com/actions/checkout#checkout-v6

(older ones will not run in near future, due to relying libraries or code runner)


nigelmegitt (approved):
LGTM, thanks.

